PR #13, opened 2022-05-25
Fragment based environment

bengioe:
Implementation of a fragment based environment + an example task from Bengio et al. (2021), binding to the sEH inhibitor.


src/gflownet/data/sampling_iterator.py
```diff
@@ -16,7 +20,7 @@ class SamplingIterator(IterableDataset):
 
     """
     def __init__(self, dataset: Dataset, model: nn.Module, batch_size: int, ctx, algo, task, device, ratio=0.5,
-                 stream=True):
+                 stream=True, log_dir: str = None):
```
pjanowski, line 23: please add the parameter to the function doc string

src/gflownet/data/sampling_iterator.py
```diff
@@ -59,6 +64,9 @@ def _idx_iterator(self):
             # Otherwise, figure out which indices correspond to this worker
             worker_info = torch.utils.data.get_worker_info()
             n = len(self.data)
+            if n == 0:
```
pjanowski, line 69: I'm curious: does this have to `yield` rather than `return`? The `__iter__` function yields but this one just returns a single numpy arrray.  Also why the need to `return` after some `yield`s?

src/gflownet/data/sampling_iterator.py
```diff
@@ -81,18 +89,21 @@ def __len__(self):
...
     def __iter__(self):
         worker_info = torch.utils.data.get_worker_info()
-        wid = (worker_info.id if worker_info is not None else 0)
-        self.rng = self.algo.rng = self.task.rng = np.random.default_rng(142857 + wid)
+        self._wid = (worker_info.id if worker_info is not None else 0)
+        self.rng = self.algo.rng = self.task.rng = np.random.default_rng(142857 + self._wid)
         self.ctx.device = self.device
+        if self.log_dir is not None:
+            os.makedirs(self.log_dir, exist_ok=True)
+            self.generated_logfile = open(f'{self.log_dir}/generated_mols_{self._wid}.csv', 'a')
```
pjanowski: I don't see this file handle getting closed. Is that intentional? Better to create the file once and then write to it using `with` context?

src/gflownet/train.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, hps: Dict[str, Any], device: torch.device):
         self.hps = {**self.default_hps(), **hps}
         self.device = device
         self.num_workers: int = self.hps.get('num_data_loader_workers', 0)
+        self.offline_ratio = 0.5
```
pjanowski, line 114: It might be good to add some comments to some of these attributes. For example, it's hard to read what the difference is between `offline_ratio` and `valid_offline_ratio`. Similarly, what is `hps`?

src/gflownet/envs/frag_mol_env.py
```diff
@@ -0,0 +1,166 @@
...
+        self.num_node_dim = len(self.frags_smi) + 1
+        self.num_edge_attr_logits = most_stems * 2
+        self.num_edge_dim = most_stems * 2
+        self.num_cond_dim = num_cond_dim
+
+        # Order in which models have to output logits
+        self.action_type_order = [GraphActionType.Stop, GraphActionType.AddNode, GraphActionType.SetEdgeAttr]
+        self.device = torch.device('cpu')
+
+    def aidx_to_GraphAction(self, g: gd.Data, action_idx: Tuple[int, int, int]):
```
pjanowski, line 54: Should `g` get a more descriptive name and should it be an optional parmeter?

src/gflownet/envs/frag_mol_env.py
```diff
@@ -0,0 +1,166 @@
...
+import torch
+import torch_geometric.data as gd
+
+from gflownet.envs.graph_building_env import Graph
+from gflownet.envs.graph_building_env import GraphAction
+from gflownet.envs.graph_building_env import GraphActionType
+from gflownet.envs.graph_building_env import GraphBuildingEnvContext
+
+
+class FragMolBuildingEnvContext(GraphBuildingEnvContext):
```
pjanowski, line 15: Please add a descriptive docstring to the class and parameter docstrings to the member functions below.

src/gflownet/models/bengio2021flow.py
```diff
@@ -0,0 +1,279 @@
...
+
+        self.lin1 = nn.Linear(dim, dim * 8)
+        self.lin2 = nn.Linear(dim * 8, num_out_per_stem)
+
+        self.set2set = Set2Set(dim, processing_steps=3)
+        self.lin3 = nn.Linear(dim * 2, num_out_per_mol)
+        self.bond2out = nn.Sequential(nn.Linear(dim * 2, dim), self.act, nn.Linear(dim, dim), self.act,
+                                      nn.Linear(dim, num_out_per_bond))
+
+    def forward(self, data, vec_data=None, do_stems=True, do_bonds=False, k=None, do_dropout=False):
```
pjanowski: Are `do_stems` and `do_bonds` used?

src/gflownet/models/bengio2021flow.py
```diff
@@ -0,0 +1,279 @@
...
+        self.num_conv_steps = num_conv_steps
+        self.dropout_rate = 0
+
+        self.act = nn.LeakyReLU()
+
+        net = nn.Sequential(nn.Linear(4, 128), self.act, nn.Linear(128, dim * dim))
+        self.conv = NNConv(dim, dim, net, aggr='mean')
+        self.gru = nn.GRU(dim, dim)
+
+        self.lin1 = nn.Linear(dim, dim * 8)
```
pjanowski: Are `lin1` and `lin2` used?

src/gflownet/models/bengio2021flow.py
```diff
@@ -0,0 +1,279 @@
...
+        'set2set.lstm.bias_hh_l0': params[19],
+        'lin3.weight': params[20],
+        'lin3.bias': params[21],
+    }
+    for k, v in param_map.items():
+        mpnn.get_parameter(k).data = torch.tensor(v)
+    return mpnn
+
+
+atomic_numbers = {
```
pjanowski: Is this full dictionary useful (only it's `len` is used)? If so would it make sense to add it to some Constants.py file or if not then just move to the top of this file for improved readability?

src/gflownet/models/graph_transformer.py
```diff
@@ -95,3 +95,45 @@ def forward(self, g: gd.Batch, cond: torch.Tensor):
             types=self.action_type_order,
         )
         return cat, self.emb2reward(graph_embeddings)
+
+
+class GraphTransformerFragGFN(nn.Module):
```
pjanowski, line 158: Please add a descriptive docstring for each of the two classes here.

src/gflownet/tasks/seh_frag.py
```diff
@@ -0,0 +1,217 @@
...
+from gflownet.train import RewardScalar
+
+
+def thermometer(v: Tensor, n_bins=50, vmin=0, vmax=1) -> Tensor:
+    bins = torch.linspace(vmin, vmax, n_bins)
+    gap = bins[1] - bins[0]
+    return (v[..., None] - bins.reshape((1,) * v.ndim + (-1,))).clamp(0, gap.item()) / gap
+
+
+class SEHTask(GFNTask):
```
pjanowski, line 49: Please add docstrings, especially to clarify the acronyms (SEH, GFN)

pjanowski: also consider marking private methods with an underscore

src/gflownet/tasks/seh_frag.py
```diff
@@ -0,0 +1,217 @@
...
+        self.opt_Z.step()
+        self.opt_Z.zero_grad()
+        self.lr_sched.step()
+        self.lr_sched_Z.step()
+        if self.sampling_tau > 0:
+            for a, b in zip(self.model.parameters(), self.sampling_model.parameters()):
+                b.data.mul_(self.sampling_tau).add_(a.data * (1 - self.sampling_tau))
+
+
+# Determined-specific code:
```
pjanowski: Should this live in gflownetR?

src/gflownet/data/sampling_iterator.py
bengioe: Since this function is a generator, I think of the `return` statements as `break`. As for why `yield` rather than constructing a list and returning it, I guess I thought it would be more elegant this way. `yield` handles the case where `stream == True` and we'd have to return an infinite iterator anyways.

src/gflownet/data/sampling_iterator.py
bengioe: The intent here is to open a file and gradually append to it as new objects are being generated by the model. What would be the right way to handle this? Reopening the log file in `'a'` mode after every minibatch? 

Note that I'm thinking of switching to `sqlite` instead, since it will scale much better and more gracefully than a csv, and presumably handle the underlying opening and closing of files better.

src/gflownet/train.py
bengioe: I added comments to clarify

src/gflownet/envs/frag_mol_env.py
bengioe: It is a necessary parameter. I added some more explicit documentation.

src/gflownet/models/bengio2021flow.py
bengioe: No, good catch. I removed much of the original code but forgot about the arguments.

src/gflownet/models/bengio2021flow.py
bengioe: True, I'll add a constant at the top of the file.

src/gflownet/tasks/seh_frag.py
bengioe: Yes, missed that.

src/gflownet/models/bengio2021flow.py
bengioe: No I removed them.

src/gflownet/data/sampling_iterator.py
pjanowski: Moving to sqlite might be a good idea. As for logging to the file, I'd do a `with open` in `a` mode directly when you are ready to write in the `log_generated` method. This ensure the file handle does get left open which is never a good idea. The other option would be to call `close()` explicitely at some point toward the end of the run but this is not a great idea: your program might fail before that, etc.

bengioe: Ok, pushed a fix

src/gflownet/tasks/seh_frag.py
bengioe: Addressed by recursionpharma/gflownetR#14.

src/gflownet/data/sampling_iterator.py
bengioe: Now addressed by #23.

pjanowski (approved):

PR #23, opened 2022-08-03
Move from csv to sqlite-based logging

bengioe:
Replace "printing to a csv file" by "inserting into a sqlite db" to log the data generated by models. This is cleaner, and it also turns to be about 20x faster when loading the data in a notebook.

Includes 2 minor fixes.

- [x] This PR depends on #13.


src/gflownet/algo/trajectory_balance.py
```diff
@@ -159,7 +159,7 @@ def not_done(lst):
                 fwd_logprob[i].append(log_probs[j].unsqueeze(0))
                 data[i]['traj'].append((graphs[i], graph_actions[j]))
                 # Check if we're done
-                if graph_actions[j].action is GraphActionType.Stop or t == self.max_len - 1:
+                if graph_actions[j].action is GraphActionType.Stop or (self.max_len and t == self.max_len - 1):
```
bearnshaw, line 162: I'm not going to prevent this, but checks like this usually mean `max_len` is not being handled well. My guess is that it is an optional parameter without a default. Could the default be some huge number instead?

src/gflownet/tasks/seh_frag.py
```diff
@@ -60,10 +60,10 @@ def __init__(self, dataset: Dataset, temperature_distribution: str, temperature_
         self.temperature_sample_dist = temperature_distribution
         self.temperature_dist_params = temperature_parameters
 
-    def _flat_reward_transform(self, y: Union[float, Tensor]) -> FlatRewards:
+    def flat_reward_transform(self, y: Union[float, Tensor]) -> FlatRewards:
```
bearnshaw, line 63: Are these `flat_reward_transform` changes needed for sqlite logging? If not, it is poor practice to slip them into a different PR. Better to separate into two PRs.

bearnshaw (approved):
@bengioe I am approving just to get caught up on PRs, but I left some comments that are worth your consideration for the future.

src/gflownet/tasks/seh_frag.py
bengioe: True, I'll be more careful next time!

src/gflownet/algo/trajectory_balance.py
bengioe: Yes that would make sense as well.

bengioe: Resolving this for now, I opened #26 as a reminder.

PR #29, opened 2022-09-29
Extract graph sampler from TB

bengioe:
This makes GraphSampler its own class instead of a method of `TrajectoryBalance`. It will be reused for other algorithms.

* adds proper masking for GraphActionCategoricals, although more functionality can still be added here.
* removes `_corrupt_actions` in `TrajectoryBalance`. `GraphSampler.sample_from_model` now instead relies on masks to take random action in the event that `GraphSampler.random_action_prob > 0`.

### DIFF
--- a/src/gflownet/algo/graph_sampling.py
+++ b/src/gflownet/algo/graph_sampling.py
@@ -1,13 +1,11 @@
 import copy
 from itertools import count
-from typing import List, Tuple
+from typing import List
 
-import numpy as np
 import torch
 from torch import Tensor
 import torch.nn as nn
 
-from gflownet.envs.graph_building_env import GraphActionCategorical
 from gflownet.envs.graph_building_env import GraphActionType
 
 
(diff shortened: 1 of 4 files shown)


bengioe:
@bearnshaw I thought it would be simpler to merge some individual components before merging the various RL baselines. This gives me a chance to review each part, add comments, make sure they make sense on their own.

bearnshaw:
Love the idea! Sorry I am only getting to the review now.

src/gflownet/algo/graph_sampling.py
```diff
@@ -0,0 +1,140 @@
...
+                else:  # If not done, try to step the self.environment
+                    gp = graphs[i]
+                    try:
+                        # self.env.step can raise AssertionError if the action is illegal
+                        gp = self.env.step(graphs[i], graph_actions[j])
+                        assert len(gp.nodes) <= self.max_nodes
+                    except AssertionError:
+                        done[i] = True
+                        data[i]['is_valid'] = False
+                        continue
```
bearnshaw, line 125: Why do you use a try except block here? It looks like all you want is to check that `len(gp.nodes) <= self.max_nodes` and if it is not, to indicate that the graph is invalid. This can be accomplished with a simple `if` statement.

bearnshaw (changes requested):

bearnshaw (approved):

src/gflownet/algo/graph_sampling.py
bengioe: `self.env.step` itself is designed to raise AssertionError for a bad action. It would probably be more clear if we raised something like a `GraphActionError`. Noted for future PRs.

PR #30, opened 2022-09-30
Improvements to GraphActionCategorial including docs and new methods

bengioe:
Adds docs and 3 methods to `GraphActionCategorial`:
- `entropy`
- `logsumexp`
- `argmax`

- [x] Depends on #29

PR #31, opened 2022-09-30
Improve validation

bengioe:
Improves metrics for multiobjective optimization

- Introduces Top-k and Tanimoto similarity reporting.
- Allows tasks to define a validation set of fixed preferences that condition the data being generated for validation
- Logs the Pareto front explicitly, including smiles of current front.

- [x] depends on #29 
- [x] depends on #30

### DIFF
--- a/src/gflownet/tasks/seh_frag_moo.py
+++ b/src/gflownet/tasks/seh_frag_moo.py
@@ -1,4 +1,8 @@
 import ast
+from collections import defaultdict
+import pathlib
+import queue
+import threading
 from typing import Any, Callable, Dict, List, Tuple, Union
 
 import numpy as np
@@ -9,6 +13,7 @@ import scipy.stats as stats
 import torch
 from torch import Tensor
 from torch.distributions.dirichlet import Dirichlet
+import torch.multiprocessing as mp
 import torch.nn as nn
 from torch.utils.data import Dataset
 import torch_geometric.data as gd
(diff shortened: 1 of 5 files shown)


src/gflownet/tasks/seh_frag_moo.py
```diff
@@ -116,78 +129,80 @@ def default_hps(self) -> Dict[str, Any]:
...
+        # possible to hack our way around that by using __subclasshook__ in our internal
+        # determined-depending code (https://www.hillelwayne.com/post/python-abc/).
+        parent = self
         try:
-            hsri_on_flat, _ = hsr_indicator.calculate_hsr(-1 * flat_rewards)
-        except Exception:
-            hsri_on_flat = 0
+            from determined.pytorch import PyTorchCallback
+        except ImportError:
+            PyTorchCallback = object
```
bearnshaw: Under what circumstances is there an import error? Old versions of determined?

Also, isn't setting to `object` brittle?

bengioe: Following up on our in-person discussion, I'm removing the `determined` dependency in a new commit.

We can do the following in our internal `determined`-facing code:
```python
def build_callbacks(self):  # Override the method in the determined Trial class
  callbacks = ParentTrainerClass.build_callbacks(self)  # Build the callbacks
  for k, v in callbacks.items():
      # This is a hack to make the callback's class inherit
      # from PyTorchCallback post-hoc:
      class CallbackSubclasser(v.__class__, PyTorchCallback):
          pass
      v.__class__ = CallbackSubclasser
  return callbacks
``` 

It's not super elegant but works.

bearnshaw (approved):

PR #39, opened 2022-12-14
Add SubTB objective + valence checks

bengioe:
This PR has two major contributions plus some housekeeping.
- Adds the SubTrajectory Balance objective [1]
- Adds valence checks to the atom environment

Housekeeping:
- Readme updates
- rewards are no longer passed around on a linear scale, but instead on a log scale (relevant when using low temperatures)
- add a FiLM-like shift & scale operation as a function of conditioning inputs to the base architecture
- reorganize how logits are generated in the model class, this is to be compatible with different environments
- remove dependency on determined in callbacks
- add `start_at_step` hyperparameter in `GFNTrainer`
- makes some computations optional in the multiobjective statistics hook
- allow keyword arguments in multiprocessing proxy

[1] [Learning GFlowNets from partial episodes for improved convergence and stability, Kanika Madan, Jarrid Rector-Brooks, Maksym Korablyov, Emmanuel Bengio, Moksh Jain, Andrei Nica, Tom Bosc, Yoshua Bengio, Nikolay Malkin](https://arxiv.org/abs/2209.12782)


bengioe:
@bearnshaw this is kind of two PRs in one. Let me know if I should split this one up in two (or more) pieces.

The build is currently failing because of a security vulnerability in a dependency, any thoughts on how to proceed?

bearnshaw:
> The build is currently failing because of a security vulnerability in a dependency, any thoughts on how to proceed?

Can you update the dependency?

bengioe:
> Can you update the dependency?

The dependency is mpmath which comes through sympy, used by MXMNet which we've used in the past as the proxy for QM9 experiments. It seems like the issue has been patched on the trunk of mpmath but no new version has been released (including on nexus). We could remove the explicit sympy dependency and raise an exception telling users to manually install sympy if they want to run QM9 experiments.

src/gflownet/algo/trajectory_balance.py
```diff
@@ -254,10 +267,45 @@ def compute_batch_losses(self, model: TrajectoryBalanceModel, batch: gd.Batch, n
...
+        # Precompute all possible subtrajectory indices that we will use for computing the loss
+        ar = torch.arange(self._subtb_max_len, device=dev)
+        tidx = [torch.tril_indices(i, i, device=dev)[1] for i in range(self._subtb_max_len)]
+        self._precomp = [
+            (torch.cat([i + tidx[T - i]
+                        for i in range(T)]),
+             torch.cat([ar[:T - i].repeat_interleave(ar[:T - i] + 1) + ar[T - i + 1:T + 1].sum()
+                        for i in range(T)]))
+            for T in range(1, self._subtb_max_len)
+        ]
```
bearnshaw: I think it would be helpful to extend the docstring and say in words what this code is doing.

src/gflownet/algo/trajectory_balance.py
```diff
@@ -254,10 +267,45 @@ def compute_batch_losses(self, model: TrajectoryBalanceModel, batch: gd.Batch, n
...
+
+    def subtb_loss_fast(self, P_F, P_B, F, R, traj_lengths):
```
bearnshaw, line 310: Same here - add a docstring that describes in words what the code is doing.

src/gflownet/envs/graph_building_env.py
```diff
@@ -173,7 +173,7 @@ def step(self, g: Graph, action: GraphAction) -> Graph:
         elif action.action is GraphActionType.SetNodeAttr:
             assert self.allow_node_attr
             assert action.source in gp.nodes
-            assert action.attr not in gp.nodes[action.source]
+            assert action.attr not in gp.nodes[action.source] or gp.nodes[action.source][action.attr] is None
```
bearnshaw, line 179: Add a comment that explains the edge case addressed here.

bearnshaw (changes requested):

bengioe:
@bearnshaw Pushed some comments to explain how the SubTB objective is computed, let me know if that clarifies things.

src/gflownet/algo/trajectory_balance.py
bearnshaw: Thanks for doing this! However, comments usually proceed the code, not follow. Would you please change the order?

bearnshaw: Looks good, thanks!

src/gflownet/envs/graph_building_env.py
bearnshaw: Looks good, thanks!

bearnshaw (approved):

PR #42, opened 2023-01-18
Add idempotency correction

bengioe:
Adds mechanisms to compensate for idempotent forward and backward actions ("different" actions that lead to the same state) and prevent under/overcounting transition probabilities.

Modifies:
- `GraphSampler` to optionally account for idempotent actions when counting backward transitions
- `TrajectoryBalance` to optionally account for idempotent actions when computing $P_F(s'|s)$ (i.e. actually computes $P_F(s'|s) = \sum_{a'\equiv a} P_F(a'|s)$ rather than just $P_F(a|s)$)
- `GraphBuildingEnv` to correctly enumerate parents of a state (taking into account isomorphic parents)

Also modifies `SamplingIterator` to rely on a clearer `'result'` key in trajectories rather than indexing.

### DIFF
--- a/src/gflownet/algo/trajectory_balance.py
+++ b/src/gflownet/algo/trajectory_balance.py
@@ -74,10 +74,12 @@ class TrajectoryBalance:
         self.length_normalize_losses = False
         self.reward_normalize_losses = False
         self.sample_temp = 1
-        self.graph_sampler = GraphSampler(ctx, env, max_len, max_nodes, rng, self.sample_temp)
-        self.graph_sampler.random_action_prob = hps['random_action_prob']
         self.is_doing_subTB = hps.get('tb_do_subtb', False)
         self.correct_idempotent = hps.get('tb_correct_idempotent', False)
+
+        self.graph_sampler = GraphSampler(ctx, env, max_len, max_nodes, rng, self.sample_temp,
+                                          correct_idempotent=self.correct_idempotent)
+        self.graph_sampler.random_action_prob = hps['random_action_prob']
         if self.is_doing_subTB:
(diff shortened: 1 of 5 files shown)


src/gflownet/algo/trajectory_balance.py
```diff
@@ -125,6 +132,47 @@ def create_training_data_from_graphs(self, graphs):
         """
         return [{'traj': generate_forward_trajectory(i)} for i in graphs]
 
+    def get_idempotent_actions(self, g: Graph, gd: gd.Data, gp: Graph, action: GraphAction):
+        """Returns the list of idempotent actions for a given transition.
+
+        Note, this is slow! Correcting for idempotency is needed to estimate p(x) correctly, but
+        isn't generally necessary if we mostly care about sampling approximately from the modes
+        of p(x).
```
bearnshaw, line 140: Love the docstring.

src/gflownet/algo/trajectory_balance.py
```diff
@@ -125,6 +132,47 @@ def create_training_data_from_graphs(self, graphs):
...
+        """
+        iaction = self.ctx.GraphAction_to_aidx(gd, action)
+        if action.action == GraphActionType.Stop:
+            return [iaction]
+        mask_name = self.ctx.action_mask_names[iaction[0]]
+        lmask = getattr(gd, mask_name)
+        nz = lmask.nonzero()
+        actions = [iaction]
+        for i in nz:
+            aidx = (iaction[0], i[0].item(), i[1].item())
```
bearnshaw, line 167: TODO (in the future): create class for this tuple with `__equals__` implementation

bearnshaw (approved):

PR #43, opened 2023-01-18
Retire redundant fragment model

bengioe:
The current `GraphTransformerGFN` supports partial action spaces, there is thus no need for a specialized `GraphTransformerFragGFN` anymore. This PR replaces the latter with the more general class.


bengioe:
Note that current build fails are fixed by #39.

bearnshaw (approved):

PR #47, opened 2023-02-03
Add pre-LayerNorm to GraphTransformer

bengioe:
Adds [pre-LayerNorm](https://arxiv.org/abs/2002.04745) to GraphTransformer and makes it the default for increased training stability.


src/gflownet/models/graph_transformer.py
```diff
@@ -104,12 +109,19 @@ def forward(self, g: gd.Batch, cond: torch.Tensor):
...
+                l_h = linear(trans(torch.cat([o, agg], 1), aug_edge_index, aug_e))
+                scale, shift = cs[:, :l_h.shape[1]], cs[:, l_h.shape[1]:]
+                o = norm1(o + l_h * scale + shift, aug_batch)
+                o = norm2(o + ff(o), aug_batch)
+            else:
+                agg = gen(norm1(o, aug_batch), aug_edge_index, aug_e)
+                l_h = linear(trans(torch.cat([o, agg], 1), aug_edge_index, aug_e))
+                scale, shift = cs[:, :l_h.shape[1]], cs[:, l_h.shape[1]:]
+                o = o + l_h * scale + shift
+                o = o + ff(norm2(o, aug_batch))
```
bearnshaw, line 125: Why is `norm2` before `ff` when `ln_type == "pre"`?

bengioe: the `pre`-LayerNorm normalizes the inputs of both modules (which are the transformer bit `gen+trans`, and the MLP bit `ff`) instead of their outputs:
image

bengioe: Actually, you made me spot an inconsistency. I was norming the input of `gen` but not `trans`! Fixed in latest commit.

bearnshaw: Meant to do that 💪

bearnshaw (approved):

PR #49, opened 2023-02-09
Speed and stability improvements to GraphActionCategorical + some tests

bengioe:
Changes log & exp based operations to clamp instead of adding an epsilon; this prevents nans that can arise from numerical instabilities (i.e. very large or small logits).

Additionally improves the `log_prob` method by changing the Python `for` loop with array operations and fancy indexing. 

Also adds tests 🎉 (the first of many)

(Also bumps version to 0.0.11)


bearnshaw (approved):

bearnshaw:
Thanks for the walkthrough last week - I understand the logic now.

PR #50, opened 2023-02-14
Add random walk features to atom-env

bengioe:
Adds:
- the ability compute random walk features of graph nodes, this is known to help GNNs' representational power (see [Dwivedi et al.](https://arxiv.org/abs/2110.07875)).
- these features by default to `MolBuildingEnvContext`, the atom environment context
- a docstring to `MolBuildingEnvContext.__init__`

### DIFF
--- a/src/gflownet/utils/graphs.py
+++ b/src/gflownet/utils/graphs.py
@@ -21,7 +21,7 @@ def random_walk_probs(g: Data, k: int, skip_odd=False):
     else:
         Pmult = P
     Pk = Pmult
-    for k in range(k):
+    for _ in range(k):
         diags.append(torch.diagonal(Pk, dim1=-2, dim2=-1))
         Pk = Pk @ Pmult
     p = torch.cat(diags, dim=0).transpose(0, 1)  # (num nodes, k)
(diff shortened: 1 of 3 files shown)


bearnshaw:
What are random walk features? Is there a reference you could share?

bengioe:
Good point, as far as I know [this is where they were introduced](https://arxiv.org/abs/2110.07875), they also made their way into the recent SoTA [GraphGPS](https://arxiv.org/abs/2205.12454) paper.

src/gflownet/utils/graphs.py
```diff
@@ -0,0 +1,28 @@
...
+    else:
+        # P = D^-1 * A
+        P = torch.diag(deg_inv) @ to_dense_adj(g.edge_index, max_num_nodes=g.num_nodes)  # (1, num nodes, num nodes)
+    diags = []
+    if skip_odd:
+        Pmult = P @ P
+    else:
+        Pmult = P
+    Pk = Pmult
+    for k in range(k):
```
bearnshaw:
```suggestion
    for _ in range(k):
```
It's more pythonic to use `_` in a dummy loop.

bearnshaw (changes requested):
Looks good, one requested change.

bearnshaw (approved):

PR #52, opened 2023-02-19
Improve fragment environment for validity, add stop_mask & original fragment stems

bengioe:
Changes the fragment environment slightly so that the agent is forced to produce valid molecules. This is done by further masking invalid actions, in particular, the agent is no longer allowed to stop unless inter-fragment bonds are labeled. 

Also further constrains the agent's choice of bonding site by using the (more restrictive) original stem attachment points of the fragments of [Bengio et al. (2021)](https://arxiv.org/abs/2106.04399). 

This means the fragment environment will be slightly less expressive, but closer to the original paper and slightly easier to train.


.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@ repos:
       - id: check-executables-have-shebangs # Ensures that (non-binary) executables have a shebang.
       - id: check-json # This hook checks json files for parseable syntax.
       - id: check-shebang-scripts-are-executable # Ensures that (non-binary) files with a shebang are executable.
-      - id: pretty-format-json # This hook sets a standard for formatting JSON files.
+# Seems unnecessary:      - id: pretty-format-json # This hook sets a standard for formatting JSON files.
```
bearnshaw: Why remove this? Is it breaking something?

src/gflownet/envs/frag_mol_env.py
```diff
@@ -28,24 +34,26 @@ def __init__(self, max_frags: int = 9, num_cond_dim: int = 0):
             The dimensionality of the observations' conditional information vector (if >0)
         """
         self.max_frags = max_frags
-        self.frags_smi = open(os.path.split(__file__)[0] + '/frags_72.txt', 'r').read().splitlines()
+        smi, stems = zip(*json.load(open(os.path.split(__file__)[0] + '/frags_72.json', 'r')))
```
bearnshaw: Does this actually close the file handle? Why not use a `with` context?

src/gflownet/envs/frags_72.json
```diff
@@ -0,0 +1 @@
...1COCCN1", [5]], ["C1COCC[NH2+]1", [5]], ["C=C", [0, 1]], ["C=C(C)C", [0]], ["C=CC", [0, 1]], ["C=N", [0]], ["C=O", [0]], ["CC", [0, 1]], ["CC(C)C", [1]], ["CC(C)O", [1]], ["CC(N)=O", [2]], ["CC=O", [1]], ["CCC", [1]], ["CCO", [1]], ["CN", [0, 1]], ["CNC", [1]], ["CNC(C)=O", [0]], ["CNC=O", [0, 2]], ["CO", [0, 1]], ["CS", [0]], ["C[NH3+]", [0]], ["C[SH2+]", [1]], ["Cl", [0]], ["F", [0]], ["FC(F)F", [1]], ["I", [0]], ["N", [0]], ["N=CN", [1]], ["NC=O", [0, 1]], ["N[SH](=O)=O", [1]], ["O", [0]], ["O=CNO", [1]], ["O=CO", [1]], ["O=C[O-]", [1]], ["O=PO", [1]], ["O=P[O-]", [1]], ["O=S=O", [1]], ["O=[NH+][O-]", [1]], ["O=[PH](O)O", [1]], ["O=[PH]([O-])O", [1]], ["O=[SH](=O)O", [1]], ["O=[SH](=O)[O-]", [1]], ["O=c1[nH]cnc2[nH]cnc12", [3, 6]], ["O=c1[nH]cnc2c1NCCN2", [8, 3]], ["O=c1cc[nH]c(=O)[nH]1", [2, 4]], ["O=c1nc2[nH]c3ccccc3nc-2c(=O)[nH]1", [8, 4, 7]], ["O=c1nccc[nH]1", [3, 6]], ["S", [0]], ["c1cc[nH+]cc1", [1, 3]], ["c1cc[nH]c1", [0, 2]], ["c1ccc2[nH]ccc2c1", [6]], ["c1ccc2ccccc2c1", [0, 2]], ["c1ccccc1", [0, 1, 2, 3, 4, 5]], ["c1ccncc1", [0, 1, 2, 4, 5]], ["c1ccsc1", [2, 4]], ["c1cn[nH]c1", [0, 1, 3, 4]], ["c1cncnc1", [0, 1, 3, 5]], ["c1cscn1", [0, 3]], ["c1ncc2nc[nH]c2n1", [2, 6]]]
```
bearnshaw: Why a file instead of a string?

bengioe: No particular reason, do you think a string would be better? I'm unaware what the preferred convention is. 

Ideally in the future, users can specify the fragments they'd want to use.

.pre-commit-config.yaml
bengioe: It was failing on the `frag_72.json` file, but I'm not sure why. JSON files aren't a central part of this code base so I thought maybe this was unnecessary.

bengioe:
Upon reflection it feels more elegant to have those fragments as a constant in the `bengio2021flow` module, which encapsulates the original task of the GFlowNet paper. This solves the JSON & file opening issues.

bearnshaw:
I agree, I like that organization better.

bearnshaw (approved):

PR #54, opened 2023-03-10
Log-reward uniformisation, logging and new hyperparams for temperature, Z-training, validation.

julienroyd:
Several changes (most of them minor):

- Added option ‘constant’ for ‘temperature_sample_dist’
- New hyperparameters:
  - objectives (variable list of objectives to optimise)
  - num_thermometer_dim (for beta encoding)
  - Z_learning_rate
  - valid_random_action_prob (now performs greedy_validation by default)
  - n_valid_prefs (number of validation preference vector to span simplex)
  - n_valid_repeats_per_pref (number of molecules to sample for each preference vector at validation)
- Added logging
  - Now saves molecules generated during validation
  - Saves hyperparameters in stdout and on disk
  - Saves ‘git_hash’ and ‘hostname’ in hyperparameters
  - To stdout during training
- Made all algorithms consistent with log-rewards


src/gflownet/algo/soft_q_learning.py
```diff
@@ -46,17 +44,17 @@ def __init__(self, env: GraphBuildingEnv, ctx: GraphBuildingEnvContext, rng: np.
         self.max_len = max_len
         self.max_nodes = max_nodes
         self.illegal_action_logreward = hps['illegal_action_logreward']
-        self.alpha = hps['sql_alpha']
+        self.alpha = hps.get(hps['sql_alpha'], 0.01)
```
bengioe: Should be `hps.get('sql_alpha', 0.01)`.

src/gflownet/tasks/qm9/qm9.py
```diff
@@ -70,21 +74,38 @@ def load_task_models(self):
...
+        return {'beta': beta, 'encoding': beta_enc}
+
+    def cond_info_to_logreward(self, cond_info: Dict[str, Tensor], flat_reward: FlatRewards) -> RewardScalar:
         if isinstance(flat_reward, list):
             flat_reward = torch.tensor(flat_reward)
-        return RewardScalar(flat_reward.flatten()**cond_info['beta'])
+        flat_reward = flat_reward.squeeze()
+        assert len(flat_reward.shape) == len(cond_info['beta'].shape), \
+            f"dangerous shape mismatch: {flat_reward.shape} vs {cond_info['beta'].shape}"
+        return RewardScalar(torch.log(flat_reward + 1e-8) * cond_info['beta'])
```
bengioe: I'd do a `.clamp` here, since it doesn't distort regular values and is not prone to NaN when going through gradients (which shouldn't be the case here, but better to use that convention throughout the code).

src/gflownet/utils/metrics.py
```diff
@@ -158,6 +149,14 @@ def sharpeRatio(p, Q, x, rf):
 
 def _sharpeRatioQPMax(p, Q, rf):
     """ Sharpe ratio maximization problem - QP formulation """
+    from cvxopt import matrix
```
bengioe, line 155: Not that it's a bad change but was there a reason for this? Non-top-level imports are unusual, might be good to leave a comment behind either way.

src/gflownet/tasks/seh_frag.py
```diff
@@ -52,25 +58,38 @@ def _load_task_models(self):
...
+        else:
+            if self.temperature_sample_dist == 'gamma':
+                loc, scale = self.temperature_dist_params
+                beta = self.rng.gamma(loc, scale, n).astype(np.float32)
+                upper_bound = stats.gamma.ppf(0.95, loc, scale=scale)
+            elif self.temperature_sample_dist == 'uniform':
+                beta = self.rng.uniform(*self.temperature_dist_params, n).astype(np.float32)
+                upper_bound = self.temperature_dist_params[1]
+            elif self.temperature_sample_dist == 'loguniform':
+                beta = np.exp(self.rng.uniform(*np.log(self.temperature_dist_params), n).astype(np.float32))
```
bengioe: mypy is having trouble following the `*` operator here unfortunately, because it cannot determine the length of the output of `np.log(..)` and finds no match for a two-argument `uniform(Number, Integer)` call. A fix is to declare the uniform bounds outside e.g.:
```
low, high = np.log(self.temperature_dist_params)
beta = np.exp(self.rng.uniform(low, high, n).astype(np.float32))
```

src/gflownet/algo/soft_q_learning.py
julienroyd: Thanks, corrected [here](https://github.com/recursionpharma/gflownet/pull/54/commits/7e6893d3d59392ca5a6bf8e89efff25d9c2e64fb)

src/gflownet/utils/metrics.py
julienroyd: Right, this shouldn't have ended up here, I moved the import inside the function since on my local environment (with M1 chip) cvxopt was really annoying to install and I ended up simply moving forward without it is only used in that one function.

I can revert it but at the same time since having the import down there removes a hard requirement which could be problematic for several users it could make sense to keep it there. What do you think?

bengioe: Makes sense, I'd recommend adding a comment explaining this and opening an issue on github so we're reminded to check back when cvxopt fixes their M1 compatibility (and so that other potential M1 users facing this problem might see it).

src/gflownet/tasks/seh_frag.py
julienroyd: Fixed [here](https://github.com/recursionpharma/gflownet/commit/1a952537df52ebd462f3d6e2d0942a8277c73868)

src/gflownet/utils/metrics.py
julienroyd: See [this issue](https://github.com/recursionpharma/gflownet/issues/56).

src/gflownet/tasks/qm9/qm9.py
julienroyd: Were you thinking about something like:
```
return RewardScalar(torch.max(torch.log(flat_reward), torch.tensor([-75.])) * cond_info['beta'])
```
or clamping before getting into log-space?

bengioe: It's safer to clamp before `log` (for 0, subnormal floats, or negative values), e.g: 

https://github.com/recursionpharma/gflownet/blob/450098e371ea2eb71c37f1f3acf64f8e65149036/src/gflownet/algo/trajectory_balance.py#L263

julienroyd: [Fixed here](https://github.com/recursionpharma/gflownet/commit/e3fdb25548395fce6598f763e391ef48d123a3c6)

bengioe:
#50 and #52 were merged, would you mind resolving conflicts and rebasing?

julienroyd:
@bengioe I merged the trunk [here](https://github.com/recursionpharma/gflownet/pull/54/commits/d96988dd69bd0c28564df05f6af02e128852dd90), we can fast-forward now. I am currently running an experiment just to validate that we still get the same results and then we should be good to close. I'll let you know when the results are in.

bengioe (approved):

julienroyd:
I am done with the validation run, results look good (see on [this link](https://github.com/recursionpharma/gflownetR/blob/3d875e36810ff307e310b95c0552c99ceb5b6d37/gflownetr/seh_frag_moo/2023_03_20_beta_PR1/analysis.ipynb)).

julienroyd:
Closing.

PR #55, opened 2023-03-13
Add parameterized backward policy

bengioe:
The main contribution of this PR is that it adds the possibility to use a parameterized backwards policy to GFN methods instead of the default uniform $1/\mathrm{num parents}$ policy.

This (also) modifies:
- `GraphTransformerGFN`, the class is made a bit more generic, which made implementing backward categorical outputs more elegant
- `GraphSampler`, so that it also adds backward actions to trajectories, and pads trajectories when relevant
- `TrajectoryBalance`, to adapt the objective computation
- `FragMolBuildingEnvContext`, to adapt the environment to output backward action masks
- `GraphBuildingEnv` and `GraphActionCategorical` to handle backward actions
- some tests to ensure some equivalence between masks and backward action counts.

Some additional minor fixes: adds a `max_nodes` hyperparameter in `SEHFrag`, removes `frags_72.txt` which is no longer needed, makes `ln_type` a parameter to `GraphTransformerGFN` (was ignored previously).

Note this still lacks some features, reserved for future PRs:
- Generate trajectories following $P_B^\theta$ when replaying terminal states
- Adapt the atom environment to also support backward policies


bearnshaw (approved):
Thanks for the in-person review!

PR #57, opened 2023-03-22
Improved validation vectors and added IGD and PC-entropy metrics

julienroyd:
Adding 3 new functions for validation and monitoring:

- `metrics.partition_hypersphere`: function for generating a predefined number of vectors which are (approximately) maximally spread-apart inside a d-dimensional hypersphere. It supports both `l1` and `l2` normalisation, which allows it to be used to either sample the space of stochastic or unit vectors. We currently use it to generate the dataset of fixed `preferences` for the validation step. The method that was previously in place generated a variable number of vectors which would grow exponentially in the number of objectives and would thus make it difficult to know how many such vectors will be generated a priori and difficult to combine with other conditioning methods such as `focus_regions`. This new partitioning strategy allows to specify the number of partitioning vectors a priori.

- `metrics.get_IGD`: implements the Inverse Generational Distance; a metric for evaluating the quality of a set of samples provided by a model optimizing multiple objectives. It takes the average across the points forming the "true" pareto-front of the distances with each of their closest generated sample in objective space. If the true pareto front is unknown, the reference set is composed of the limiting points of a discretised unit-hypercube. It seeks to capture the coverage of the true pareto front -- a value of 0 means that each point of the true pareto front are covered by at least one generated sample. 

- `metrics.get_PC_entropy`: computes the entropy of the Pareto-Clustered sample distribution; we attribute a "label" to each generated sample corresponding to the point in the "true" pareto front that is the closest. We call this Pareto-clustering (clustering the samples w.r.t the points on the pareto front). These counts form a categorical distribution and its entropy captures how uniformly the samples are distributed in terms of proximity to each of the points on the pareto front. If the true pareto front is unknown, the limiting points of the unit hypercube are used as described for the computation of IGD. The maximum value is the entropy of a k-dimensional categorical distribution, where k is the number of points forming the true pareto front.

### DIFF
--- a/src/gflownet/tasks/seh_frag_moo.py
+++ b/src/gflownet/tasks/seh_frag_moo.py
@@ -203,7 +203,8 @@ class SEHMOOFragTrainer(SEHFragTrainer):
                                temperature_sample_dist=self.hps['temperature_sample_dist'],
                                temperature_parameters=self.hps['temperature_dist_params'],
                                num_thermometer_dim=self.hps['num_thermometer_dim'], wrap_model=self._wrap_model_mp)
-        self.sampling_hooks.append(MultiObjectiveStatsHook(256, self.hps['log_dir'], compute_reach=True))
+        self.sampling_hooks.append(
+            MultiObjectiveStatsHook(256, self.hps['log_dir'], compute_igd=True, compute_pc_entropy=True))
 
         n_obj = len(self.hps['objectives'])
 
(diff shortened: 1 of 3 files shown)


julienroyd:
Converted as Draft for now. I want to:

- Add some links to documentation (reach-metric actually corresponds to a known metric in the literature)
- Add some derivatives of this metric including a clustering-based version that would account for how evenly populated each of the reference-points neighbourhoods are.
- Add a parameter to `reach_metric` to dilate the range of variations for well performing empirical fronts.

julienroyd:
@bengioe ready for review.

- Find the analysis on the chosen metrics [here](https://github.com/recursionpharma/gflownetR/blob/1082607f4a74094054e1a68c76ef55c61e163fcc/gflownetr/seh_frag_moo/sandbox/pareto_metrics.ipynb)
- Find a validation experiment at the head of this branch [here](https://github.com/recursionpharma/gflownetR/blob/d6b0872c16e853980fbad549ff56ebb9237c3dc8/gflownetr/seh_frag_moo/2023_03_30_beta_PR2/analysis.ipynb)

bengioe (approved):
Looks good to me.

julienroyd:
Closing.

PR #59, opened 2023-04-04
Reduce load on shared memory

bengioe:
We've been seeing some "too many files open" errors in trunk in some edge cases. This PR introduces two changes to alleviate that:

- set the `prefetch_factor` of DataLoaders to 1 instead of the default, 2.
- give the option to pickle the messages between `SamplingIterator` and `MPModelProxy`, `mp_pickle_messages`.

These changes reduce the number of temporarily created shared memory files (one per tensor!). Surprisingly, I thought that pickling messages would slow things down, but with a quick rerun of `seh_frag_moo` I found the following:
- 8 workers, pickled messages, 2.05 it/s
- 8 workers, normal messages, 1.48 it/s
- no workers, (therefore no messages), 0.45 it/s

This is rather confusing, and may not transfer to different envs, batch sizes, etc, but may be worth investigating a bit more.


bearnshaw (approved):
I'm approving this, but pickle has fallen out of favor for a variety of reasons. I don't know what is the preferred alternative these days, might be worth looking into.

PR #60, opened 2023-04-05
Fix edge attributes in FragMolBuildingEnvContext

### LINKED ISSUE #58: Interpretation of edge_attr in FragMolBuildingEnvContext
Hi!

Thanks for the great work on GFlowNets. I am going through the code, to understand and reproduce the multi-objective molecule generation experiments (from the Multi-objective GFlowNets paper).

I am a little bit confused by the interpretation of `edge_attributes` created by `FragMolBuildingEnvContext.graph_to_Data` 

The `edge_attr` tensor is initialized as
```
edge_attr = torch.zeros((len(g.edges) * 2, self.num_edge_dim))
```

and then filled as follows
```
for i, e in enumerate(g.edges):
    ad = g.edges[e]
    a, b = e
    for n, offset in zip(e, [0, self.num_stem_acts]):
        idx = ad.get(f'{int(n)}_attach', -1) + 1
        edge_attr[i * 2, idx] = 1
        edge_attr[i * 2 + 1, idx] = 1
        ...
```

I do not completely understand how I should interpret the features `edge_attr[i * 2, :]` and `edge_attr[i * 2 + 1, :]` of the edge number `i`. 

If I understand correctly, in the code above `idx` takes value 0 when the attribute `f'{a}_attach'` (`f'{b}_attach'`) is not set. Otherwise, `idx` is the 1-based index of the stem of node a (b). In particular, if both attributes `f'{a}_attach'` and `f'{b}_attach'` are set to 0, then the `edge_attr[i * 2, :] = [0, 1, 0, 0, 0, ... ]`.

Questions:
- Is `edge_attr[2 * i, :]` supposed to be the one-hot encoding of the stems of nodes a and b connected by the i-th edge?
- If that's the case. Shouldn't the code be something like this?
```

self.num_edge_dim = (most_stems + 1) * 2

...

edge_attr = torch.zeros((len(g.edges) * 2, self.num_edge_dim))

...

for i, e in enumerate(g.edges):
    ad = g.edges[e]
    a, b = e
    for n, offset in zip(e, [0, self.num_stem_acts]):
        idx = ad.get(f'{int(n)}_attach', -1) + 1
        edge_attr[i * 2, idx + offset] = 1
        edge_attr[i * 2 + 1, idx + offset] = 1
        ...
```

timgaripov:
Resolves #58 following @bengioe's suggestion 

Changes:
- change `num_edge_dim` to `(most_stems + 1) * 2` since one-hot features encode `most_stems + 1` options (no attribute, or one of at most `most_stems` attributes) per node
- fill `edge_attr` using proper offsets
- change `num_edge_attr_logits` to `most_stems * 2` since `set_edge_attr_mask` values at positions `>= 2 * most_stems` are always set to zero (`attach_point + self.num_stem_acts < most_stems * 2`)

### DIFF
--- a/src/gflownet/envs/frag_mol_env.py
+++ b/src/gflownet/envs/frag_mol_env.py
@@ -54,8 +54,8 @@ class FragMolBuildingEnvContext(GraphBuildingEnvContext):
         self.num_new_node_values = len(self.frags_smi)
         self.num_node_attr_logits = 0
         self.num_node_dim = len(self.frags_smi) + 1
-        self.num_edge_attr_logits = (most_stems + 1) * 2
-        self.num_edge_dim = most_stems * 2
+        self.num_edge_attr_logits = most_stems * 2
+        self.num_edge_dim = (most_stems + 1) * 2
         self.num_cond_dim = num_cond_dim
         self.edges_are_duplicated = True
         self.edges_are_unordered = False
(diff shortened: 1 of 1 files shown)


bengioe:
I ran a short comparison with trunk on seh_frag and, good news, there appears to be a small but significant improvement in performance.

image

image

bengioe (approved):

timgaripov:
Thanks for running the experiment! Glad that the fix helps.

PR #62, opened 2023-04-10
Add flow matching objective

bengioe:
Adds the flow matching objective, which is good to have for reproducibility and completeness. Will also be used in toy environment.

- [x] Depends on #55 and #59.


bearnshaw (approved):
Reviewed during meeting, looks ready.

PR #64, opened 2023-04-19
Fix numerical stability of logsumexp

bengioe:
This is minor but may bite us in edge cases, the `logsoftmax` operation in `GraphActionCategorical` was essentially computing `log(exp(logits - b))` which is terrible for the numerical stability of the gradient. This PR is just replacing `log(exp(logits - b))` by `logits - b`.

Another edge case: this PR also adds default outputs for the max scatter reduction. This avoids the max value being 0 if there is nothing of a certain type and all other max values are less than zero 0.

This PR also adds tests that would have failed before it, and some repeated operations in `GraphActionCategorial` methods are now done with `_compute_batchwise_max`.


aebrer (approved):
looks good to me from my limited understanding

PR #65, opened 2023-04-21
Add special case for nitrogen valence

bengioe:
This PR corrects the max valence of nitrogen, and prevents the agent from forming too many bonds with it(that would induce a molecule considered invalid by RDKit).

The fix is definitely a heuristic, but it is compatible with the nitrogen bond variety found in ChembL. Consider the following snippet:
```python
import pandas as pd
import tqdm
import multiprocessing
from rdkit import Chem
from gflownet.envs.mol_building_env import MolBuildingEnvContext

chembl = pd.read_parquet('path/to/chembl.parquet')
ctx = MolBuildingEnvContext(['Br', 'C', 'Cl', 'F', 'N', 'O', 'P', 'S'])

def f(smi):
    mol = Chem.MolFromSmiles(smi)
    if mol is None:
        return []
    g = ctx.obj_to_graph(mol)
    bts = []
    for i in g.nodes:
        if g.nodes[i]['v'] == 'N':
            ns = [g.edges[(i, u)] for u in g.neighbors(i)]
            bt = (len(ns),
                  tuple(sorted([b.get('type',Chem.BondType.SINGLE).real for b in ns])),
                  g.nodes[i].get('charge', 0),
                  g.nodes[i].get('no_impl', 0))
            bts.append(bt)
    return bts

def g():
    bonds = []
    with multiprocessing.Pool(64) as pool:
        for i in tqdm.tqdm(pool.imap_unordered(f, chembl.index), total=len(chembl)):
            bonds += i
    print(set(bonds))
            
if __name__ == '__main__':
    g()
```

which outputs:
```python
{(0, (), 0, 0),
 (1, (1,), 0, 0),
 (1, (2,), -1, 0),
 (1, (2,), -1, True),
 (1, (2,), 0, 0),
 (1, (3,), 0, 0),
 (2, (1, 1), -1, True),
 (2, (1, 1), 0, 0),
 (2, (1, 1), 0, True),
 (2, (1, 2), 0, 0),
 (2, (1, 2), 0, True),
 (2, (1, 3), 1, True),
 (2, (2, 2), 1, 0),
 (2, (2, 2), 1, True),
 (3, (1, 1, 1), 0, 0),
 (3, (1, 1, 1), 0, True),
 (3, (1, 1, 1), 1, True),
 (3, (1, 1, 2), 1, 0),
 (3, (1, 1, 2), 1, True),
 (4, (1, 1, 1, 1), 1, 0),
 (4, (1, 1, 1, 1), 1, True)}
```

Seeing this suggested the following heuristic to me: limit Nitrogen's valence to 3 unless it is charged, in which case we make it 5. The positive charge reduces the maximum number of bonds to 4, which is compatible with the above.


src/gflownet/data/sampling_iterator.py
```diff
@@ -209,14 +209,16 @@ def __iter__(self):
...
+                self.log_generated(trajs[num_offline:], rewards[num_offline:], flat_rewards[num_offline:], {
+                    k: v[num_offline:] for k, v in cond_info.items()
+                })
             if num_online > 0:
                 extra_info = {}
                 for hook in self.log_hooks:
                     extra_info.update(
-                        hook(trajs[num_offline:], rewards[num_offline:], flat_rewards[num_offline:],
-                             {k: v[num_offline:] for k, v in cond_info.items()}))
+                        hook(trajs[num_offline:], rewards[num_offline:], flat_rewards[num_offline:], {
```
bengioe: This change and the above has nothing to do with this PR, but coincides with a `yapf` version change, which incidentally is making the `python/library` pipeline fail... 🤔

bengioe:
Seems I epic failed the blacken+merge step on git so closing this PR and reopening, see #76.

Issue #44, opened 2023-01-18
Add a MANIFEST.in


bengioe:
Currently non-`.py` files are not bundled in the wheel (e.g. when doing a `pip install git+http://.../gflownet.git`), which means some files would be missing (like `gflownet/envs/frag_72.txt`). This can be fixed by adding a `MANIFEST.in` file.

PR #66, opened 2023-04-21
chore: paint it black and sand ruff edges

dmaljovec:
# What?
 - Run `black` against this repository
 - Fix `ruff .` errors
 - Configure `tox` and `pre-commit` to use `black` in place of `yapf` and `ruff` in place of `flake8`

# Why?
 - Our internal company standard has recently been switched to `black`

PR #67, opened 2023-04-24
Fix version source in setup.py

bengioe:
Fix `setup.py` to read from the `VERSION` file. This is necessary since we got rid of `setup.cfg` in #66.


bengioe:
Fixes #44 by adding `MANIFEST.in`.

PR #74, opened 2023-05-02
feat: get semantic versioning working with standard build

dmaljovec:
# What?
 - Omit patch version from `VERSION` file (let the build infer it from last known matching major.minor release)
 - Remove `setuptools-scm` from build requires (we are not going to use this after all)
 - Defer requirements gathering to non-dynamic specification within `pyproject.toml`
 - Leverage the `SEMVER` variable within `setup.py` and fallback to inspecting the local git directory (this will fail on a source build likely if it tries to reconstruct the version without any git context, but people should be using the wheel as this is a pure python codebase)

# Why?
- These should be minimal steps to get this compatible with our internal `python/library` pipeline.
- I have attempted to add parenthetical statements to the whats above to describe the need for each.


bengioe (approved):

dmaljovec:
Let's see if it works!

dmaljovec:
The failure on `trunk` is a result of the build not being flexible enough. I am going to fix that up and re-run the build against the fixed pipeline.

dmaljovec:
Fixed: https://github.com/recursionpharma/gflownet/commit/116d616ec8956478fd68ea47a4788089dd43ced4 🎉

PR #75, opened 2023-05-03
Add thermometer encoding for the preference vector

bengioe:
Adds the option to use thermometer encoding for the preference vector in `seh_frag_moo`, makes this option the default (due to apparent increased performance, plots to come). Preferences were previously just encoded as their `[0,1]` value.

Side fix: Removes redundant creation of `self.task` in `seh_frag_moo`.


bengioe:
Plots confirm improvement on `seh_frag_moo`
image

julienroyd:
Looks good to me! 

Only thing: maybe I would set the default to `False` since we are having some doubts that it consistently helps but it does increase the memory use.

julienroyd (approved):

PR #78, opened 2023-05-10
Disable gradient computations in multiprocessing proxy

bengioe:
Disables gradient computations in multiprocessing proxy, since they cannot be passed between processes anyways. This saves some amount of memory when running with CUDA.

Also makes error raising more robust by ensuring errors make their way back to the callee instead of silently killing the thread.


bearnshaw (approved):
Reviewed during meeting, looks good.

PR #79, opened 2023-05-10
Add checkpoint_every hyperparameter

bengioe:
Adds a `checkpoint_every`  hyperparameter to the main training loop, which defaults to `validate_every`.

### DIFF
--- a/src/gflownet/train.py
+++ b/src/gflownet/train.py
@@ -239,6 +239,7 @@ class GFNTrainer:
         self.sampling_model.to(self.device)
         epoch_length = max(len(self.training_data), 1)
         valid_freq = self.hps.get("validate_every", 0)
+        # If checkpoint_every is not specified, checkpoint at every validation epoch
         ckpt_freq = self.hps.get("checkpoint_every", valid_freq)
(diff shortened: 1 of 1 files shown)


src/gflownet/train.py
```diff
@@ -238,6 +238,8 @@ def run(self, logger=None):
         self.model.to(self.device)
         self.sampling_model.to(self.device)
         epoch_length = max(len(self.training_data), 1)
+        valid_freq = self.hps.get("validate_every", 0)
+        ckpt_freq = self.hps.get("checkpoint_every", valid_freq)
```
aebrer, line 243: What does passing in the `valid_freq` here do?

aebrer (approved):
left one comment for my benefit, otherwise looks good

src/gflownet/train.py
bengioe: It makes the checkpoint frequency default to the value of `validate_every` (`hps` is a `dict`, `a_dict.get(key, default)` returns `default` when `key not in a_dict`). I'll add a comment.

PR #80, opened 2023-05-10
Add max_nodes and max_edges to MolBuildingEnvContext

bengioe:
Adds `max_nodes` and `max_edges` to MolBuildingEnvContext. Also checks molecules' sanity slightly differently, by making sure RDKit is happy with the SMILES of the molecule.

### DIFF
--- a/src/gflownet/envs/mol_building_env.py
+++ b/src/gflownet/envs/mol_building_env.py
@@ -356,7 +356,6 @@ class MolBuildingEnvContext(GraphBuildingEnvContext):
     def is_sane(self, g: Graph) -> bool:
         try:
             mol = self.graph_to_mol(g)
-            assert mol is not None
         except Exception:
             return False
         if mol is None:
--- a/src/gflownet/envs/mol_building_env.py
+++ b/src/gflownet/envs/mol_building_env.py
@@ -355,7 +355,7 @@ class MolBuildingEnvContext(GraphBuildingEnvContext):
 
     def is_sane(self, g: Graph) -> bool:
         try:
-            mol = self.graph_to_obj(g)
+            mol = self.graph_to_mol(g)
             assert mol is not None
         except Exception:
             return False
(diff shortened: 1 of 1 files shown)


src/gflownet/envs/mol_building_env.py
```diff
@@ -333,12 +348,15 @@ def graph_to_mol(self, g: Graph) -> Mol:
             mp.AddBond(e[0], e[1], d.get("type", BondType.SINGLE))
         mp.CommitBatchEdit()
         Chem.SanitizeMol(mp)
-        return mp
+        # Not sure why, but this seems to find errors that SanitizeMol doesn't, and it saves us trouble downstream,
+        # since MolFromSmiles returns None if the SMILES encoding of the molecule is invalid, which seemingly happens
+        # occasionally (although very rarely) even if RDKit is able to create a SMILES string from `mp`.
+        return Chem.MolFromSmiles(Chem.MolToSmiles(mp))
```
aebrer, line 354: Praise: I can confirm that this is basically best practice in my experience.

src/gflownet/envs/mol_building_env.py
```diff
@@ -333,12 +348,15 @@ def graph_to_mol(self, g: Graph) -> Mol:
...
 
     def is_sane(self, g: Graph) -> bool:
         try:
             mol = self.graph_to_mol(g)
-            assert Chem.MolFromSmiles(Chem.MolToSmiles(mol)) is not None
+            assert mol is not None
         except Exception:
             return False
         if mol is None:
```
aebrer, line 361: suggestion: is this code still needed? it seems safe to delete since you are already asserting that `mol is not None`

aebrer (approved):

src/gflownet/envs/mol_building_env.py
bengioe: Good point!

PR #81, opened 2023-05-11
Fix: missing offline trajectory data in TB

bengioe:
Fixes `create_training_data_from_graphs` in TB, add missing `bck_logprobs` and `result` keys.


bearnshaw (approved):
Reviewed during meeting, looks good.

PR #85, opened 2023-05-16
Make PRs trigger github workflows

bengioe:
This change should make it possible for external pull requests to [trigger our github workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-events-for-forked-repositories).

It seems like we may still need to [trigger them manually](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) for Recursion-external forks, which sounds like a good thing, but at least it will be possible.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -1,5 +1,9 @@
 name: Build-and-Test
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - trunk
+  pull_request:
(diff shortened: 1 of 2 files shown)


bengioe:
Seems like now every workflow is ran twice, investigating.

bengioe:
Seems like this is the right combo, awaiting review.

dmaljovec (approved):

PR #88, opened 2023-05-20
add GitPython

SobhanMP:
the dependency is used in seh_frag_moo to store the git version, would be convenient to install all of the deps in one shot.


SobhanMP:
nvm, i jus saw it was already there... :sweat_smile:

Issue #61, opened 2023-04-06
Potential modification of flat_rewards after batch-creation


julienroyd:
Experiments using a ReplayBuffer have lead to surprising results where the model would be incapable to learn if our list of `flat_rewards` (tensors of shape (2,) ) would be stacked into a tensor of shape (batch, 2) before being pushed in the buffer. We were able to fix this issue by creating a copy of each item pushed into the replay buffer (note: only copying `flat_rewards` may have been required).

- Code state with the bug ([commit 6a1b7b0](https://github.com/recursionpharma/gflownet/commit/6a1b7d062e8ca7ec729ca8202c3f7950f18af055))
- Fix ([commit 8e1163a](https://github.com/recursionpharma/gflownet/commit/8e1163a4fbac9adfc28e286569382386dc8e738d))

Our hypothesis is that `flat_rewards` (and potentially other tensors) are modified after the batch is created, which was not harmful when discarding the batch after the parameter update (on-policy) but became harmful when this modification caused the `flat_rewards` in the replay buffer to also be modified which would cause the model to train on wrong trajectory-reward pairs when this trajectory would be re-sampled from the buffer later on. It could be worth identifying what operation caused the buffer data to change to validate that this operation is intentional and should indeed occur.

bengioe:
Seems to have been fixed by #121 . We were doing in-place modification through sliced tensors (which I didn't think was possible!) of the original data rather than creating new tensors.

PR #89, opened 2023-05-24
Adding support for goal-conditioning, replay-buffer and upgrading to torch1.13 

julienroyd:
Several changes relating to Multi-Objective tasks:

* Adding support for goal-conditioning. The type of goals currently supported are _focus-regions_ (hypercones in objective space), defined by a focus-direction and a `focus_cosim` threshold. The sampling distribution of these focus-regions can be fixed, `hyperspherical`, or `learned-tabular`, see `hps.focus_type`.
* Adding support for a simple `ReplayBuffer`. Currently storing full trajectories. As of now it completely replaces online-sampling when `use_replay_buffer=True`.
* Extends the use of multiprocessing-proxy to any python object (not just torch models).
* Adds hyperparameter `num_final_gen_steps` to save trajectories from the final distribution before exiting (useful for analysis).
* Upgrading to `torch==1.13.1+cu113`


src/gflownet/data/sampling_iterator.py
```diff
@@ -201,49 +218,97 @@ def __iter__(self):
...
+            #  TODO: implement that per-task (in case they don't apply the same beta and log transformations)
             rewards = torch.exp(log_rewards / cond_info["beta"])
-
             if num_online > 0 and self.log_dir is not None:
                 self.log_generated(
-                    trajs[num_offline:],
-                    rewards[num_offline:],
-                    flat_rewards[num_offline:],
-                    {k: v[num_offline:] for k, v in cond_info.items()},
+                    deepcopy(trajs[num_offline:]),
```
bengioe, line 238: Were those to temporarily try guarding against replay buffer shenanigans? Or do things break if we remove the `deepcopy`s?

julienroyd: I had a different (but most probably related) issue with monitoring; the focus_accuracy metric was wrong probably because `flat_rewards` was getting modified somehow, I didn't take the time to thoroughly investigate but there is an issue opened [here](https://github.com/recursionpharma/gflownet/issues/61) and my guess is that solving it would allow us to remove deepcopy's eveywhere in sampling_iterator (both for the replay buffer and the metrics).

I also documented this [here](https://docs.google.com/presentation/d/1VpnorJ1SIJQR54vO1ZnOoKpfGY8ysF_lvl0zQmPIhxo/edit?usp=sharing), see slide on April 10th.

src/gflownet/data/sampling_iterator.py
```diff
@@ -143,23 +156,29 @@ def __iter__(self):
...
                 )
                 graphs = [self.ctx.mol_to_graph(m) for m in mols]
                 trajs = self.algo.create_training_data_from_graphs(graphs)
-                num_online = self.online_batch_size
+
             else:  # If we're not sampling the conditionals, then the idcs refer to listed preferences
                 num_online = num_offline
                 num_offline = 0
-                cond_info = self.task.encode_conditional_information(torch.stack([self.data[i] for i in idcs]))
+                cond_info = self.task.encode_conditional_information(
```
bengioe, line 175: In the original MOO code, the argument here was validation preferences (used in scalarization). Feels like this would prevent running a scalarization MOO validation epoch? Or does it still work because we assume preferences are first in the encoding tensor?
(`preferences = steer_info[:, : len(self.objectives)].float()` later) 
Would be good to document either way.

src/gflownet/data/sampling_iterator.py
```diff
@@ -201,49 +218,97 @@ def __iter__(self):
...
                         )
                     )
-                batch.extra_info = extra_info
+
+            if self.replay_buffer is not None:
+                # If we have a replay buffer, we push the online trajectories in it
+                # and resample immediately such that the "online" data in the batch
+                # comes from a more stable distribution (try to avoid forgetting)
+                # Important: note that the 'online' metrics now will be describing the data
+                # in the replay buffer and not purely the data from the last generated batch
```
bengioe: Is this still the case? You seem to be calling `log_generated` before overwriting the trajectories.

src/gflownet/data/sampling_iterator.py
```diff
@@ -201,49 +218,97 @@ def __iter__(self):
...
+                # Relabels some of the online trajectories with hindsight
+                assert hasattr(
+                    self.task, "relabel_condinfo_and_logrewards"
+                ), "Hindsight requires the task to implement relabel_condinfo_and_logrewards"
+                # samples indexes of trajectories without repeats
+                hindsight_idxs = torch.randperm(num_online)[: int(num_online * self.hindsight_ratio)] + num_offline
+                cond_info, log_rewards = self.task.relabel_condinfo_and_logrewards(
+                    cond_info, log_rewards, flat_rewards, hindsight_idxs
+                )
+                log_rewards[torch.logical_not(is_valid)] = self.algo.illegal_action_logreward
```
bengioe, line 296: Setting invalid rewards feels like it should be outside the `if self.hindsight...`?

src/gflownet/tasks/seh_frag.py
```diff
@@ -205,6 +219,16 @@ def setup(self):
             "none": (lambda x: None),
         }[hps["clip_grad_type"]]
 
+        # saving hyperparameters
+        git_hash = git.Repo(__file__, search_parent_directories=True).head.object.hexsha[:7]
+        self.hps["gflownet_git_hash"] = git_hash
+
+        os.makedirs(self.hps["log_dir"], exist_ok=True)
+        fmt_hps = "\n".join([f"{f'{k}':40}:\t{f'({type(v).__name__})':10}\t{v}" for k, v in self.hps.items()])
```
bengioe: Small QoL suggestion: use `sorted(self.hps.items())`

src/gflownet/train.py
```diff
@@ -98,6 +99,7 @@ def __init__(self, hps: Dict[str, Any], device: torch.device):
         self.model: nn.Module
         # `sampling_model` is used by the data workers to sample new objects from the model. Can be
         # the same as `model`.
+        self.replay_buffer: ReplayBuffer
```
bengioe: Breaks the `sampling_model` comment above it.

src/gflownet/train.py
```diff
@@ -243,15 +278,21 @@ def run(self, logger=None):
         ckpt_freq = self.hps.get("checkpoint_every", valid_freq)
         train_dl = self.build_training_data_loader()
         valid_dl = self.build_validation_data_loader()
+        final_dl = self.build_final_data_loader()
```
bengioe: Maybe add `if num_final_gen_steps > 0:`?

src/gflownet/utils/focus_model.py
```diff
@@ -0,0 +1,103 @@
...
+class TabularFocusModel(FocusModel):
+    """
+    Tabular model of the feasibility of focus directions for goal-condtioning.
+    We keep a count of the number of times each focus direction has been sampled and whether
+    this direction succesfully lead to a sample in this region of the objective space. The (unormalized) likelihood
+    of a focus direction being feasible is then given by the ratio of these numbers.
+    If a focus direction has not been sampled yet it obtains the maximum likelihood of one.
+    """
+
+    def __init__(self, device, n_objectives: int, state_space_res: int, focus_cosim: float) -> None:
```
bengioe: Would be good to document the arguments here and in methods below.

src/gflownet/utils/focus_model.py
```diff
@@ -0,0 +1,103 @@
...
+            count += idx_increments
+
+    def sample_focus_directions(self, n):
+        """
+        Samples n focus directions from the focus model.
+        """
+        sampling_likelihoods = torch.zeros_like(self.focus_dir_count).float().to(self.device)
+        sampling_likelihoods[self.focus_dir_count == 0] = 1.0
+        sampling_likelihoods[torch.logical_and(self.focus_dir_count > 0, self.focus_dir_population_count > 0)] = 1.0
+        sampling_likelihoods[torch.logical_and(self.focus_dir_count > 0, self.focus_dir_population_count == 0)] = 0.1
```
bengioe: This `0.1` looks like a hyperparameter, would be good to at least be tied to an attribute of the class.

bengioe (changes requested):
Left minor comments here and there, otherwise LGTM!

src/gflownet/data/sampling_iterator.py
julienroyd: Yeah maybe a bit opaque right now but we simply concatenate the preferences and focus-directions together in the validation dataset. This is in part because preference-conditioning and goal-conditioning are not an either-or, we support using both of them together in some cases (e.g. fixed set of goals with dirichlet-sampled preferences). I'll add some documentation in the validation dataset constructor and in encode_conditional_information() to specify that we make these positional assumptions.

src/gflownet/data/sampling_iterator.py
julienroyd: Oh yes good catch, a good example of "comments lie"! I had written that in the first iteration and didn't clean it up when I fixed it.

src/gflownet/data/sampling_iterator.py
julienroyd: So this is done twice:
- The first time is [here](https://github.com/recursionpharma/gflownet/blob/358901016d7e85662995a90cb9cc77b78823c9cb/src/gflownet/data/sampling_iterator.py#L225)
- The second time is after re-labeling some rewards using hindsight.

My understanding was that invalid trajectories correspond to implausible molecules or ill-constructed graphs, and that they should remain invalid no matter which goal was assigned to them. So I re-apply the `is_valid` mask such that if some invalid trajectories were picked up by hindsight and re-attributed a reward, that reward would not overwrite the invalid tag that they had been assigned.

Does that makes sense?

bengioe: Ah right, this is guarding against the task potentially giving a non-zero reward to invalid trajectories. Makes sense.

src/gflownet/utils/focus_model.py
julienroyd: Makes sense, I created a new attribute for it to make it more explicit. I haven't wired it in the hps as I don't think it makes a big difference for now, but it could be advisable to add it eventually in the future in a case where the number of infeasible directions would largely outnumber the number of feasible directions (e.g. in massively multi-objective cases).

bengioe (approved):

PR #92, opened 2023-05-27
add inv_logp_prob and tree-backup

SobhanMP:
I'll add some tests later.


src/gflownet/envs/graph_building_env.py
```diff
@@ -521,6 +521,12 @@ def to(self, device):
             self.masks = [i.to(device) for i in self.masks]
         return self
 
+    def duplicate(self):
```
bengioe: To follow torch naming conventions, this would be better named `clone`.

PR #93, opened 2023-06-06
Feature: better configs

bengioe:
The repository is now at a size where it makes no sense to have arbitrary strings used as hyperparameter names.   
This PR replaces the "giant-dict-as-config" with a structured & typed config, which makes each module responsible for defining & documenting its own configuration.

Tasks:

- [x] fix tests to use new config
- [x] finish writing default configuration docstrings

Because this is a "breaking" change this PR updates the minor version to 1, making this release `0.1.0`.


pyproject.toml
```diff
@@ -75,6 +75,7 @@ dependencies = [
     "botorch",
     "pyro-ppl",
     "gpytorch",
+    "foliconf",
```
bearnshaw: Why use your home-rolled `foliconf` when there are so many config management tools out there?

src/gflownet/utils/metrics.py
```diff
@@ -33,7 +33,7 @@ def compute_focus_coef(
     assert (
         focus_limit_coef > 0.0 and focus_limit_coef <= 1.0
     ), f"focus_limit_coef must be in (0, 1], now {focus_limit_coef}"
-    focus_gamma_param = np.log(focus_limit_coef) / np.log(focus_cosim)
+    focus_gamma_param = torch.tensor(np.log(focus_limit_coef) / np.log(focus_cosim)).float()
```
bearnshaw: This change is not like the others, did you intend to include it in this PR?

tests/test_frag_env.py
bearnshaw, line 1: These are the only changes needed in the tests?

bengioe: Yes, that's probably a sign that we need more tests.

pyproject.toml
bengioe: My reasoning was that
- I've used this mini-library in the past for personal projects and it's been helpful and very easy to setup
- It's very easy to switch from it to OmegaConf (which IMO induces more complexity), and some work but straightforward to switch from it to hydra (which induces even more complexity)
- It has some neat aspects that other code-defined config management tools don't have, like lazy loading (useful when lots of modules having configurations but not all of them need to be imported).

I'm happy to switch to OmegaConf or something similar if it makes more sense.

src/gflownet/utils/metrics.py
bengioe: I think I needed it to pass test, but you're right it's out of place; I'll investigate.

pyproject.toml
bengioe: Considering practices at Valence I did the switch to omegaconf ;)

src/gflownet/utils/metrics.py
bengioe: Reverted.

pyproject.toml
```diff
@@ -75,6 +75,7 @@ dependencies = [
     "botorch",
     "pyro-ppl",
     "gpytorch",
+    "omegaconf",
```
maclandrol: [minor]: any strong opinion on using omegaconf  ?  (also might make sense to specify versions).

src/gflownet/algo/advantage_actor_critic.py
```diff
@@ -38,29 +35,25 @@ def __init__(
...
         self.env = env
         self.rng = rng
-        self.max_len = max_len
-        self.max_nodes = max_nodes
-        self.illegal_action_logreward = hps["illegal_action_logreward"]
-        self.entropy_coef = hps.get("a2c_entropy", 0.01)
-        self.gamma = hps.get("a2c_gamma", 1)
-        self.invalid_penalty = hps.get("a2c_penalty", -10)
+        self.max_len = cfg.algo.max_len
+        self.max_nodes = cfg.algo.max_nodes
```
maclandrol, line 46: could result in ConfigAttributeError if missing some keys in the middle.

src/gflownet/data/sampling_iterator.py
```diff
@@ -28,7 +29,7 @@ def __init__(
         self,
         dataset: Dataset,
         model: nn.Module,
-        batch_size: int,
+        cfg: Config,
         ctx,
         algo,
         task,
```
maclandrol, line 33: Is there a reason why some arguments are parameterized by the config, while others (ratio, stream, etc) need to be provided to the class ?

src/gflownet/algo/trajectory_balance.py
```diff
@@ -28,29 +30,21 @@ def logZ(self, cond_info: Tensor) -> Tensor:
         raise NotImplementedError()
 
 
-class TrajectoryBalance:
+class TrajectoryBalance(GFNAlgorithm):
     """ """
```
maclandrol: empty doc string

src/gflownet/data/sampling_iterator.py
maclandrol: doc string was not updated.

src/gflownet/tasks/config.py
```diff
@@ -0,0 +1,91 @@
...
+    objectives: List[str] = field(default_factory=lambda: ["seh", "qed", "sa", "wt"])
+
+
+@dataclass
+class QM9TaskConfig:
+    # TODO: a proper class for temperature-conditional sampling
+    temperature_sample_dist: str = "uniform"
+    temperature_dist_params: List[Any] = field(default_factory=lambda: [0.5, 32])
+    num_thermometer_dim: int = 32
+    h5_path = "/data/chem/qm9/qm9.h5"
```
maclandrol: Might make sense to not hardcode this.

src/gflownet/tasks/qm9/qm9.py
```diff
@@ -127,79 +129,73 @@ def compute_flat_rewards(self, mols: List[RDMol]) -> Tuple[FlatRewards, Tensor]:
...
-        self.mb_size = hps["global_batch_size"]
-        self.clip_grad_param = hps["clip_grad_param"]
+        self.mb_size = self.cfg.algo.global_batch_size
+        self.clip_grad_param = self.cfg.opt.clip_grad_param
         self.clip_grad_callback = {
             "value": (lambda params: torch.nn.utils.clip_grad_value_(params, self.clip_grad_param)),
             "norm": (lambda params: torch.nn.utils.clip_grad_norm_(params, self.clip_grad_param)),
             "none": (lambda x: None),
-        }[hps["clip_grad_type"]]
+        }[self.cfg.opt.clip_grad_type]
```
maclandrol, line 198: consider breaking this into two statements.

src/gflownet/train.py
```diff
@@ -100,21 +105,23 @@ def __init__(self, hps: Dict[str, Any], device: torch.device):
...
         self.ctx: GraphBuildingEnvContext
         self.task: GFNTask
         self.algo: GFNAlgorithm
 
-        # Override default hyperparameters with the constructor arguments
-        self.hps = {**self.default_hps(), **hps}
+        # There are three sources of config values
+        #   - The default values specified in individual config classes
+        #   - The default values specified in the `default_hps` method, typically what is defined by a task
+        #   - The values passed in the constructor, typically what is called by the used
```
maclandrol: by the `user` ?

src/gflownet/train.py
```diff
@@ -100,21 +105,23 @@ def __init__(self, hps: Dict[str, Any], device: torch.device):
...
+        # The final config is obtained by merging the three sources
+        self.cfg = OmegaConf.structured(Config())
+        self.set_default_hps(self.cfg)
+        OmegaConf.merge(self.cfg, hps)
```
maclandrol, line 122: Given the complexity of the config, I would suggest to consider adding a linter, just to make sure that user provided configs are compliant to the expected structure (avoid missing key error, overwriting of default, etc)

maclandrol (commented):
First past, see comments.

src/gflownet/algo/advantage_actor_critic.py
```diff
@@ -38,29 +35,25 @@ def __init__(
...
         rng: np.random.RandomState
             rng used to take random actions
-        hps: Dict[str, Any]
-            Hyperparameter dictionary, see above for used keys.
-        max_len: int
-            If not None, ends trajectories of more than max_len steps.
-        max_nodes: int
-            If not None, ends trajectories of graphs with more than max_nodes steps (illegal action).
+        cfg: Config
+            The experiment configuration
```
hadim, line 39: Since the type are in the arguments nowadays, there is no need to duplicate those in the docstring (any doc generation tool can read the type from the signature of the function/method).

hadim: That would require quite some time to edit all your docstring but just sharing FYI moving forward.

src/gflownet/config.py
```diff
@@ -0,0 +1,94 @@
...
+    lr_decay: float = 20_000
+    weight_decay: float = 1e-8
+    momentum: float = 0.9
+    clip_grad_type: str = "norm"
+    clip_grad_param: float = 10.0
+    adam_eps: float = 1e-8
+
+
+@dataclass
+class Config:
```
hadim, line 47: If you ever need to save/load a config, you could add some utility methods such as `to_{dict|yaml|json}`, `from_{dict|yaml|json}`

src/gflownet/tasks/config.py
```diff
@@ -0,0 +1,91 @@
...
+        - "gamma": gamma distribution
+        - "constant": constant temperature
+    temperature_dist_params : List[Any]
+        The parameters of the temperature distribution. E.g. for the "uniform" distribution, this is the range.
+    num_thermometer_dim : int
+        The number of thermometer encoding dimensions to use.
+    """
+
+    # TODO: a proper class for temperature-conditional sampling
+    temperature_sample_dist: str = "uniform"
```
hadim, line 25: If not any string is allowed you can have something like `Literal["uniform", "method2", "method3"]` (or enum).

src/gflownet/tasks/qm9/qm9.py
hadim: Why `(lambda x: None)` instead of `lambda x: None`?

src/gflownet/tasks/seh_frag.py
```diff
@@ -116,83 +118,82 @@ def compute_flat_rewards(self, mols: List[RDMol]) -> Tuple[FlatRewards, Tensor]:
...
     def setup_env_context(self):
         self.ctx = FragMolBuildingEnvContext(
-            max_frags=self.hps["max_nodes"], num_cond_dim=self.hps["num_thermometer_dim"]
+            max_frags=self.cfg.algo.max_nodes, num_cond_dim=self.cfg.task.seh.num_thermometer_dim
         )
 
     def setup(self):
-        hps = self.hps
         RDLogger.DisableLog("rdApp.*")
         self.rng = np.random.default_rng(142857)
```
hadim, line 190: I noticed you often pass a rng object. I have the habit of passing random see instead (from which I init an rng object). Any rational for that? Mine is that I always like and prefer to have "serializable" objects in my signatures.

I am just wondering.

src/gflownet/tasks/seh_frag.py
```diff
@@ -116,83 +118,82 @@ def compute_flat_rewards(self, mols: List[RDMol]) -> Tuple[FlatRewards, Tensor]:
...
 
     def setup_env_context(self):
         self.ctx = FragMolBuildingEnvContext(
-            max_frags=self.hps["max_nodes"], num_cond_dim=self.hps["num_thermometer_dim"]
+            max_frags=self.cfg.algo.max_nodes, num_cond_dim=self.cfg.task.seh.num_thermometer_dim
         )
 
     def setup(self):
-        hps = self.hps
         RDLogger.DisableLog("rdApp.*")
```
hadim, line 189: With datamol, you can use `with dm.without_rdkit_log()` and/or `dm.disable_rdkit_log()` :-D

hadim: See https://github.com/datamol-io/datamol/blob/main/datamol/log.py

src/gflownet/tasks/seh_frag.py
```diff
@@ -203,40 +204,40 @@ def setup(self):
...
-        os.makedirs(self.hps["log_dir"], exist_ok=True)
-        fmt_hps = "\n".join([f"{f'{k}':40}:\t{f'({type(v).__name__})':10}\t{v}" for k, v in sorted(self.hps.items())])
-        print(f"\n\nHyperparameters:\n{'-'*50}\n{fmt_hps}\n{'-'*50}\n\n")
-        with open(pathlib.Path(self.hps["log_dir"]) / "hps.json", "w") as f:
-            json.dump(self.hps, f)
+        os.makedirs(self.cfg.log_dir, exist_ok=True)
+        print("\n\nHyperparameters:\n")
+        yaml = OmegaConf.to_yaml(self.cfg)
+        print(yaml)
+        with open(pathlib.Path(self.cfg.log_dir) / "hps.yaml", "w") as f:
```
hadim, line 239: FYI: we almost everywhere use a lib called `fsspec` (https://github.com/fsspec/filesystem_spec). In short, it replaces `open()` with `fsspec.open` and allows you to work with local AND non-local files (HTTPS, SSH, S3, GCS, etc).

We also have a module in datamol that wrap some basic functions on top of `fsspec`: https://github.com/datamol-io/datamol/blob/main/datamol/utils/fs.py

src/gflownet/tasks/seh_frag.py
```diff
@@ -203,40 +204,40 @@ def setup(self):
...
 
-        os.makedirs(self.hps["log_dir"], exist_ok=True)
-        fmt_hps = "\n".join([f"{f'{k}':40}:\t{f'({type(v).__name__})':10}\t{v}" for k, v in sorted(self.hps.items())])
-        print(f"\n\nHyperparameters:\n{'-'*50}\n{fmt_hps}\n{'-'*50}\n\n")
-        with open(pathlib.Path(self.hps["log_dir"]) / "hps.json", "w") as f:
-            json.dump(self.hps, f)
+        os.makedirs(self.cfg.log_dir, exist_ok=True)
+        print("\n\nHyperparameters:\n")
+        yaml = OmegaConf.to_yaml(self.cfg)
+        print(yaml)
```
hadim, line 238: FYI: we often use a logging library called `loguru` (https://github.com/Delgan/loguru).

hadim (approved):
See a few comments below. Most of them are FYI mostly!

Nice refactoring and I guess it will make config management much easier. I also appreciate the nice and explicit variable naming (something that is underestimated IMO xD).

src/gflownet/tasks/seh_frag.py
bengioe: I'm leaving this for a future PR, but I'd like to improve the way both the rngs and device objects are used and passed around in the code, which right now is messy because I'm dealing with pytorch multiprocessing workers in a hacky way.

pyproject.toml
bengioe: Re:versions, I think the "Recursion way" has been to be as unspecific as possible in `pyproject` but to update the `requirements/*` files every month, and then if something breaks with latest version to pin it in `pyproject`.

hadim: We have been following the same strategy (just using conda instead of pip). In short, our "library `env.yml` files" contain _mostly_ lower bounds or no bounds and for any specific project (internal, research, client, etc), we were using specific `env.yml` files with more specific bounds (including equal and upper bounds).

I think it's a nice setup, so you can quickly catch any incompat with the latest versions of your upstream deps. On that point, be aware that it's super easy to schedule a weekly CI run for your tests, so even when not working on a lib you know whether it's still compatible with the latest upstream versions. See this as an example: https://github.com/datamol-io/datamol/blob/fc0431b0d62da3d160338f887649374f51c35d22/.github/workflows/test.yml#L11-L12

src/gflownet/tasks/seh_frag.py
maclandrol: On fsspec, I have also started using https://github.com/fsspec/universal_pathlib and it's quite nice.

pyproject.toml
maclandrol: Noted, as hadim said, we use similar principles, my comment was more about lower bounds, since in `omegaconf`,   `gpytorch` and `botorch` often have breaking changes.

src/gflownet/data/sampling_iterator.py
bengioe: Good point; user may want to instantiate multiple `SamplingIterator` instances with different batch_sizes, rolled back.

src/gflownet/tasks/config.py
bengioe: Good point, also noticed I was missing the conversion function.

src/gflownet/tasks/qm9/qm9.py
bengioe: I think that was inherited from `yapf`, but we're using `black` now.

src/gflownet/algo/advantage_actor_critic.py
bengioe: Actually I'm noticing my IDE is just generating the docstring with the type, so there's no extra typing involved, but yeah it does look redundant.

PR #96, opened 2023-06-16
Recursive decompose routine for fragment env

bengioe:
This PR should allow use to express back molecules into fragment graphs (provided the molecule can be expressed as such a graph).


bengioe:
Folded into #110, closing.

PR #98, opened 2023-07-06
Refactor tasks

bengioe:
This PR refactors tasks into reusable components that will make it easier to define new tasks and methods.

The PR is pretty big, but a lot of it is moving code around (or deleting repeated code):

- Creates a notion of `Conditional`, implements `TemperatureConditional`, `MultiObjectiveWeightedPreferences`, `FocusRegionConditional`
- Separates the commonalities between `seh_frag` and `seh_frag_moo` into a more generic `StandardOnlineTrainer` that is meant to be easily subclassed for new tasks.
- Adds some implementation notes and comments
- Adds a `validate_batch` routine that's useful for debugging, e.g. new environments and datasets

Also fixes some bugs:
- Makes `valid_offline_ratio` a flag and sets it explicitly in tasks where it wasn't properly set
- `first_graph_idx` was incorrectly calculated in SubTB (affected logging of logZ values)
- QM9Dataset was returning the wrong shape for its rewards
- Adds a `allow_5_valence_nitrogen` flag to `MolBuildingEnvContext`, this is needed in some cases, see `tasks/qm9.py`.
- Adds an explicit `stop_mask` to `MolBuildingEnvContext.graph_to_Data`
- Fixes incorrect default objective name in `seh_frag_moo`
- Fixes the default configurations in the tasks' `main` that hadn't been updated
- Fixes a number of routines where `focus_cond` was assumed to exist (but we can now turn it off).


src/gflownet/data/sampling_iterator.py
```diff
@@ -318,9 +323,36 @@ def __iter__(self):
...
+            # Only activate for debugging your environment or dataset (e.g. the dataset could be
+            # generating trajectories with illegal actions)
+            if self.do_validate_batch:
+                self.validate_batch(batch, trajs)
+
             self.train_it += worker_info.num_workers if worker_info is not None else 1
             yield batch
 
+    def validate_batch(self, batch, trajs):
+        for actions, atypes in [
```
maclandrol: The loop here does not seem to be useful.

src/gflownet/envs/mol_building_env.py
```diff
@@ -231,9 +234,10 @@ def graph_to_Data(self, g: Graph) -> gd.Data:
             max_atom_valence = self._max_atom_valence[ad.get("fill_wildcard", None) or ad["v"]]
             # Special rule for Nitrogen
             if ad["v"] == "N" and ad.get("charge", 0) == 1:
-                # This is definitely a heuristic, but to keep things simple we'll limit Nitrogen's valence to 3 (as
+                # This is definitely a heuristic, but to keep things simple we'll limit* Nitrogen's valence to 3 (as
```
maclandrol, line 237: [minor]: What if it's negatively charged or already has +2 charge ?. There is also some edge cases around radical electrons.

src/gflownet/tasks/config.py
```diff
@@ -72,16 +48,13 @@ class SEHMOOTaskConfig:
     max_train_it: Optional[int] = None
     n_valid: int = 15
     n_valid_repeats: int = 128
-    objectives: List[str] = field(default_factory=lambda: ["seh", "qed", "sa", "wt"])
+    objectives: List[str] = field(default_factory=lambda: ["seh", "qed", "sa", "mw"])
```
hadim, line 51: Just curious why do you need a factory and lambda versus simply: ` objectives: List[str] = ["seh", "qed", "sa", "mw"]`?

hadim (approved):
Not much to say!

Just a simple curiosity question below.

src/gflownet/models/graph_transformer.py
```diff
@@ -145,6 +145,27 @@ class GraphTransformerGFN(nn.Module):
...
+    _action_type_to_graph_part = {
+        GraphActionType.Stop: "graph",
+        GraphActionType.AddNode: "node",
+        GraphActionType.SetNodeAttr: "node",
+        GraphActionType.AddEdge: "non_edge",
+        GraphActionType.SetEdgeAttr: "edge",
+        GraphActionType.RemoveNode: "node",
+        GraphActionType.RemoveNodeAttr: "node",
+        GraphActionType.RemoveEdge: "edge",
+        GraphActionType.RemoveEdgeAttr: "edge",
```
maclandrol, line 159: question: if i understand properly, removing edge attribute is equivalent to setting bond type to SINGLE ?

src/gflownet/tasks/config.py
maclandrol: ^ unless it has changed recently, i believe you cannot have mutable variables as default.

src/gflownet/tasks/config.py
```diff
@@ -1,30 +1,10 @@
 from dataclasses import dataclass, field
-from typing import Any, List, Optional, Tuple
+from typing import List, Optional, Tuple
 
 
 @dataclass
 class SEHTaskConfig:
```
maclandrol, line 6: if this still needed in that case ? (for backward compat ?)

src/gflownet/tasks/qm9/qm9.py
```diff
@@ -139,86 +99,48 @@ def set_default_hps(self, cfg: Config):
...
-            weight_decay=self.cfg.opt.weight_decay,
-            eps=self.cfg.opt.adam_eps,
-        )
-        self.opt_Z = torch.optim.Adam(Z_params, self.cfg.opt.learning_rate, (0.9, 0.999))
-        self.lr_sched = torch.optim.lr_scheduler.LambdaLR(self.opt, lambda steps: 2 ** (-steps / self.cfg.opt.lr_decay))
-        self.lr_sched_Z = torch.optim.lr_scheduler.LambdaLR(
-            self.opt_Z, lambda steps: 2 ** (-steps / self.cfg.opt.lr_decay)
+    def setup_env_context(self):
+        self.ctx = MolBuildingEnvContext(
+            ["H", "C", "N", "F", "O"], expl_H_range=[0, 1, 2, 3], num_cond_dim=32, allow_5_valence_nitrogen=True
```
maclandrol: are hydrogen atoms explicitly added in the dataset already (or are 3D coordinates required in the task) ? If not, having hydrogen addition as explicit action `H` might not be required.

src/gflownet/tasks/qm9/qm9.py
```diff
@@ -139,86 +99,48 @@ def set_default_hps(self, cfg: Config):
...
 
 def main():
-    """Example of how this model can be run outside of Determined"""
+    """Example of how this model can be run."""
     yaml = YAML(typ="safe", pure=True)
     config_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), "qm9.yaml")
     with open(config_file, "r") as f:
         hps = yaml.load(f)
-    trial = QM9GapTrainer(hps, torch.device("cpu"))
+    trial = QM9GapTrainer(hps, torch.device("cuda"))
```
maclandrol: would make sense to check cuda_availability first.

src/gflownet/trainer.py
```diff
@@ -119,13 +121,12 @@ def __init__(self, hps: Dict[str, Any], device: torch.device):
...
         self.set_default_hps(self.cfg)
-        OmegaConf.merge(self.cfg, hps)
+        # OmegaConf returns a fancy object but we can still pretend it's a Config instance
+        self.cfg = OmegaConf.merge(self.cfg, hps)  # type: ignore
 
         self.device = device
-        # idem, but from `self.test_data` during validation.
-        self.valid_offline_ratio = 1
         # Print the loss every `self.print_every` iterations
-        self.print_every = 1000
```
maclandrol, line 128: Was this set for debugging, then you forgot, or the desired default every step is now `1`

maclandrol (approved):
See comments.

src/gflownet/tasks/config.py
bengioe: Yeah correct: https://stackoverflow.com/questions/69564830/python-dataclass-setting-default-list-with-values

src/gflownet/envs/mol_building_env.py
bengioe: I've been meaning to write down all these design decisions in a doc and have them approved/improved by an actual [comp]chemist. Or is this work that's been done somewhere that I could take inspiration from?

src/gflownet/data/sampling_iterator.py
bengioe: Good catch, I'll add a proper check for backward actions.

src/gflownet/tasks/config.py
hadim: Oki thanks!

src/gflownet/tasks/qm9/qm9.py
bengioe: Ah you're right the `H` action is not needed here.

src/gflownet/trainer.py
bengioe: Yes, actually I'll make that a setting for future debugging.

src/gflownet/models/graph_transformer.py
bengioe: For the atom environment that's the only edge attribute, so yes correct.

src/gflownet/tasks/qm9/qm9.py
bengioe: Is it standard to check for cuda and fallback to cpu? I feel like if I was expecting my code to run on cuda and it silently didn't, I'd rather the experiment fail with an exception than silently take 50x longer.

maclandrol: It's for making it configurable so someone without access to a GPU would still be able to run it (this could make sense too in case there are integration test to run on github). 

But you are right that failing explicitly is better.

src/gflownet/envs/mol_building_env.py
maclandrol: No work that I am aware of internally. Happy to help in reviewing the design doc. Note that for Nitrogen atoms in aromatic rings, there are some additional consideration, when doing Graph --> Mol later (some info here: https://github.com/datamol-io/datamol/blob/fc0431b0d62da3d160338f887649374f51c35d22/datamol/_sanifix4.py#L61) aft

src/gflownet/algo/trajectory_balance.py
```diff
@@ -363,7 +363,7 @@ def compute_batch_losses(
             traj_losses = self.subtb_loss_fast(log_p_F, log_p_B, per_graph_out[:, 0], clip_log_R, batch.traj_lens)
             # The position of the first graph of each trajectory
             first_graph_idx = torch.zeros_like(batch.traj_lens)
-            first_graph_idx = torch.cumsum(batch.traj_lens[:-1], 0, out=first_graph_idx[1:])
+            torch.cumsum(batch.traj_lens[:-1], 0, out=first_graph_idx[1:])
```
SobhanMP, line 366: is this a typo?

bengioe: No actually the previous version was an error (which only affected the printed values for logZ)
```
# I was doing
b = [1,2,3]
a = [0,0,0]
a = cumsum(b[:-1]) 
# now = a = [1, 3] because that's what cumsum returns
# but doing
a = [0,0,0]
cumsum(b[:-1], out=a[1:]) 
# is the same as
a[1:] = cumsum(b[:-1])
# now a is [0, 1, 3]
```

src/gflownet/tasks/qm9/qm9.py
bengioe: Made device configurable 👍

src/gflownet/tasks/config.py
bengioe: I thought I should leave it for consistency

src/gflownet/algo/trajectory_balance.py
SobhanMP: oh, i see

PR #100, opened 2023-07-17
Add support for backwards actions in MolBuildingEnvContext

bengioe:
This PR:
- adds support for backward actions in `MolBuildingEnvContext`,
- adds tests for `MolBuildingEnvContext` that check that backwards masks are correct,
- adds a toy atom environment where the reward is simply the number of rings in a carbon-only molecule of up to 6 atoms.
- fixes a bug in `GraphBuildingEnv.parent`, whereby 1-node graphs with attributes would have their parents misenumerated (which hadn't manifested itself because until now we didn't have contexts with more than 1 node attribute).

### DIFF
--- a/src/gflownet/envs/mol_building_env.py
+++ b/src/gflownet/envs/mol_building_env.py
@@ -8,7 +8,13 @@ import torch_geometric.data as gd
 from rdkit.Chem import Mol
 from rdkit.Chem.rdchem import BondType, ChiralType
 
-from gflownet.envs.graph_building_env import Graph, GraphAction, GraphActionType, GraphBuildingEnvContext
+from gflownet.envs.graph_building_env import (
+    Graph,
+    GraphAction,
+    GraphActionType,
+    GraphBuildingEnvContext,
+    graph_without_edge,
+)
 from gflownet.utils.graphs import random_walk_probs
 
 DEFAULT_CHIRAL_TYPES = [ChiralType.CHI_UNSPECIFIED, ChiralType.CHI_TETRAHEDRAL_CW, ChiralType.CHI_TETRAHEDRAL_CCW]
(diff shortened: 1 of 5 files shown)


src/gflownet/envs/mol_building_env.py
```diff
@@ -77,19 +83,22 @@ def __init__(
...
-            for k, s, e in zip(self.atom_attrs, [0] + list(np.cumsum(num_atom_logits)), np.cumsum(num_atom_logits))
+            for k, s, e in zip(
+                self.settable_atom_attrs, [0] + list(np.cumsum(num_atom_logits)), np.cumsum(num_atom_logits)
+            )
         }
         # The attribute and value each logit dimension maps back to
         self.atom_attr_logit_map = [
             (k, v)
-            for k in self.atom_attrs
-            if k != "v"
```
maclandrol, line 92: could you explain what this changes ? What was the `v` key ?

src/gflownet/envs/mol_building_env.py
```diff
@@ -167,9 +185,28 @@ def aidx_to_GraphAction(self, g: gd.Data, action_idx: Tuple[int, int, int], fwd:
...
+            attr = self.bond_attrs[act_col]
+            return GraphAction(t, source=a.item(), target=b.item(), attr=attr)
 
     def GraphAction_to_aidx(self, g: gd.Data, action: GraphAction) -> Tuple[int, int, int]:
         """Translate a GraphAction to an index tuple"""
+        for u in [self.action_type_order, self.bck_action_type_order]:
+            if action.action in u:
+                type_idx = u.index(action.action)
+                break
+        else:
```
maclandrol, line 210: is this likely to happen ? (an action not in with fwd or bck action list ?

src/gflownet/envs/mol_building_env.py
```diff
@@ -201,7 +238,22 @@ def GraphAction_to_aidx(self, g: gd.Data, action: GraphAction) -> Tuple[int, int
...
-        type_idx = self.action_type_order.index(action.action)
+        elif action.action is GraphActionType.RemoveNode:
+            row = action.source
+            col = 0
+        elif action.action is GraphActionType.RemoveNodeAttr:
+            row = action.source
+            col = self.settable_atom_attrs.index(action.attr)
+        elif action.action is GraphActionType.RemoveEdge:
+            row = ((g.edge_index.T == torch.tensor([(action.source, action.target)])).prod(1)).argmax()
+            row = int(row) // 2  # edges are duplicated, but edge logits are not
```
maclandrol: I  don't really get this line `row = int(row) // 2`

doesn't line 248 give you the index of the directed edge to remove ? Why take `row // 2` again here ?
Also on the above, you are adding two edges per bond right ? and removing both simultaneously or one at a time ?

src/gflownet/envs/mol_building_env.py
bengioe: `'v'` is set by the `AddNode` action, i.e. during the node's creation, rather than the `SetNodeAttr` action. Here I'm calculating which logits are going to be needed to parameterize the `SetNodeAttr` action, but since `v` is already set by construction, there's no need to have logits for it. I was previously excluding it manually, instead I made a `settable_atom_attrs` list that can't contain `v` and reuse it in the relevant places.

src/gflownet/envs/mol_building_env.py
bengioe: I try to future-proof these kind of methods, in the past as I've extended the codebase to add new actions I had some silent errors like this.

src/gflownet/envs/mol_building_env.py
bengioe: torch geometric assumes all edges are directed, so to have an undirected graph we must encode the edge list as `edge_index = [[a,b], [b,a], [c,d], [d,c], ...]`. If we have `n` (undirected) edges, we only ask the model to produce `n` logits, which is why in `aidx_to_GraphAction` we do `act_row * 2` to retrieve the edge endpoints in `g.edge_index`.
So yes in some sense we remove both simultaneously, from the point of view of torch geometric.

maclandrol: Yes, I understand that part. I was confused by the `*2` since there would need to be a specific ordering in the `edge_index` matrix to fetch the right couple of nodes. I am assuming from your example, that for any bond, the two corresponding directed edges `(u,v), (v,u)` will always be contiguous in `edge_index`.

maclandrol (approved):

src/gflownet/envs/mol_building_env.py
bengioe: Ah good point, yes it is the case that they are contiguous. I'll update the comments.

PR #103, opened 2023-08-17
Faster molecule encoding

bengioe:
This modifies `MolBuildingEnvContext.graph_to_Data` by:
- replacing torch arrays with numpy arrays (rationale: doing `a[idx] = v` is twice as fast with numpy for some reason)
- replacing the rather dumb "remove an edge and test if graph is disconnected" test by an enumeration of bridges, provided by networkx for much less computational cost.
- removing the default random walk features computation (they existed for testing purposes and are not strictly necessary)
- replacing the creation of the complement graph using networkx by a simpler enumeration of all missing edges (for the `AddEdge` action)

I'm also sneaking a dependency update in this PR, `torch-geometric 2.0.3->2.3.1`, because the `2.0.3` version seems incompatible with the latest `torch==1.13.1` build.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,10 +57,7 @@ requires-python = ">=3.8,<3.10"
 dynamic = ["version"]
 dependencies = [
     "torch==1.13.1",
-    # These pins are specific on purpose, some of these packages have
-    # unstable APIs since they are fairly new. We could instead pin
-    # them as >= in dev until something breaks?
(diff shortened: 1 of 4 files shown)


maclandrol (commented):
Nothing to comment. Quite surprising that there is such a performance gap between `np` and `torch` for that operation.

maclandrol (approved):

PR #105, opened 2023-08-21
Adding detailed balance

bengioe:
Short PR, since DB is extremely close to TB/SubTB, which adds the detailed balance (first proposed in [1]).

This also:
- makes the used variant (TB/SubTB/DB) an Enum
- makes `do_length_normalize` a proper flag since it seems potentially important for DB
- Fixes the incorrect default `device` flag in some scripts

Experiments are still running to check that everything is ok.

[1] Bengio, Y., Lahlou, S., Deleu, T., Hu, E., Tiwari, M., and Bengio, E.; GFlowNet foundations

### DIFF
--- a/src/gflownet/algo/config.py
+++ b/src/gflownet/algo/config.py
@@ -1,5 +1,12 @@
 from dataclasses import dataclass
 from typing import Optional
+from enum import Enum
+
+
+class TBVariant(Enum):
+    TB = 0
+    SubTB1 = 1
+    DB = 2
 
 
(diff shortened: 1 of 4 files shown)


bengioe:
I'm satisfied with DB's performance here, although considering accounts from other colleagues it could be much higher (there was no HP tuning here, and we are not using the commonly used transition-replay method).
image
image
LN = length-normalized losses; uPB = uniform P_B; pPB = parameterized P_B

PR #106, opened 2023-08-23
Support for LR sequence environments

bengioe:
This PR adds support for left-to-right sequence generation with GFNs.

PR changes:
- Adds a `model_is_autoregressive` option in trajectory balance, which avoids unnecessary computations in the AR case
- Simplifies `SamplingIterator` somewhat by letting contexts defined how they want their objects logged
- Adds an optional `slice_dict` parameter to `GraphActionCategorical` (useful to reuse this class for sequences)
- Adds `SeqBuildingEnv` and `AutoregressiveSeqBuildingContext`, which implement LR sequence generation MDPs
- Adds a standard `SeqTransformerGFN` as a sequence model
- Adds a `ToySeqTask` to show how to use the introduced features (the task itself is to produce sequences with doubled letters in them).

### DIFF
--- a/src/gflownet/envs/graph_building_env.py
+++ b/src/gflownet/envs/graph_building_env.py
@@ -3,7 +3,7 @@ import enum
 import re
 from collections import defaultdict
 from functools import cached_property
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import networkx as nx
 import numpy as np
@@ -423,10 +423,11 @@ class GraphActionCategorical:
         self,
         graphs: gd.Batch,
         logits: List[torch.Tensor],
-        keys: List[str],
+        keys: List[Union[str, None]],
         types: List[GraphActionType],
         deduplicate_edge_index=True,
         masks: List[torch.Tensor] = None,
+        slice_dict: Optional[dict[str, torch.Tensor]] = None,
     ):
         """A multi-type Categorical compatible with generating structured actions.
 
@@ -461,13 +462,16 @@ class GraphActionCategorical:
             be graph-level (i.e. if there are `k` graphs in the Batch
             object, this logit tensor would have shape `(k, m)`)
         types: List[GraphActionType]
-           The action type each logit corresponds to.
+            The action type each logit corresponds to.
         deduplicate_edge_index: bool, default=True
-           If true, this means that the 'edge_index' keys have been reduced
-           by e_i[::2] (presumably because the graphs are undirected)
+            If true, this means that the 'edge_index' keys have been reduced
+            by e_i[::2] (presumably because the graphs are undirected)
         masks: List[Tensor], default=None
-           If not None, a list of broadcastable tensors that multiplicatively
-           mask out logits of invalid actions
+            If not None, a list of broadcastable tensors that multiplicatively
+            mask out logits of invalid actions
+        slice_dist: Optional[dict[str, Tensor]], default=None
+            If not None, a map of tensors that indicate the start (and end) the graph index
+            of each object keyed. If None, uses the `_slice_dict` attribute of the graphs.
         """
         self.num_graphs = graphs.num_graphs
         assert all([i.ndim == 2 for i in logits])
@@ -504,9 +508,9 @@ class GraphActionCategorical:
             for k in keys
         ]
         # This is the cumulative sum (prefixed by 0) of N[i]s
+        slice_dict = graphs._slice_dict if slice_dict is None else slice_dict
         self.slice = [
-            graphs._slice_dict[k].to(dev) if k is not None else torch.arange(graphs.num_graphs + 1, device=dev)
-            for k in keys
+            slice_dict[k].to(dev) if k is not None else torch.arange(graphs.num_graphs + 1, device=dev) for k in keys
         ]
         self.logprobs = None
 
--- a/src/gflownet/envs/graph_building_env.py
+++ b/src/gflownet/envs/graph_building_env.py
@@ -781,6 +781,7 @@ class GraphBuildingEnvContext:
     """A context class defines what the graphs are, how they map to and from data"""
 
     device: torch.device
+    action_type_order: List[GraphActionType]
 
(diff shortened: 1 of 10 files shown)


src/gflownet/envs/graph_building_env.py
```diff
@@ -874,3 +879,7 @@ def mol_to_graph(self, mol: Mol) -> Graph:
             The corresponding Graph representation of that molecule.
         """
         raise NotImplementedError()
+
+    def object_to_log_repr(self, g: Graph) -> str:
```
maclandrol, line 884: is this on purpose ? Could still make sense to have a representation of the graph here

src/gflownet/envs/mol_building_env.py
```diff
@@ -457,3 +457,7 @@ def is_sane(self, g: Graph) -> bool:
         if mol is None:
             return False
         return True
+
+    def object_to_log_repr(self, g: Graph):
```
maclandrol, line 463: This assumes that `graph_to_mol` always return a valid mol. Is there a check done somewhere before ?

src/gflownet/envs/seq_building_env.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(self, variant):
+        super().__init__()
+
+    def new(self):
+        return Seq()
+
+    def step(self, g: Graph, a: GraphAction):
+        s: Seq = deepcopy(g)  # type: ignore
+        if a.action == GraphActionType.AddNode:
```
maclandrol, line 41: only sequential addition is supported at this moment ?  Is it on a token level or the vocab include various k-mers ?

src/gflownet/envs/seq_building_env.py
```diff
@@ -0,0 +1,133 @@
...
+            s.seq.append(a.value)
+        return s
+
+    def count_backward_transitions(self, g: Graph, check_idempotent: bool = False):
+        return 1
+
+    def parents(self, g: Graph):
+        s: Seq = deepcopy(g)  # type: ignore
```
maclandrol, line 49: [ignore] there is a lot of deep copy action to not modify the object, doesn't that impact performance a bit ?

src/gflownet/envs/seq_building_env.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class SeqBatch:
+    def __init__(self, seqs: List[torch.Tensor], pad: int):
+        self.seqs = seqs
+        self.x = pad_sequence(seqs, batch_first=False, padding_value=pad)
+        self.mask = self.x.eq(pad).T
+        self.lens = torch.tensor([len(i) for i in seqs], dtype=torch.long)
+        # This tells where (in the flattened array of outputs) the non-masked outputs are.
+        # E.g. if the batch is [["ABC", "VWXYZ"]], logit_idx would be [0, 1, 2, 5, 6, 7, 8, 9]
```
maclandrol, line 67: [minor] do you have an explicit `<eos>` token in your vocab ? Since the sequences are flattened and concatenated into a larger sequence, it might be useful to explicitely learn what define the end of a sequence (or something similar for the begining of sequence, well, any sequence separator really)

src/gflownet/envs/seq_building_env.py
```diff
@@ -0,0 +1,133 @@
...
+    """This class masquerades as a GraphBuildingEnvContext, but actually generates sequences of tokens.
+
+    This context gets an agent to generate sequences of tokens from left to right, i.e. in an autoregressive fashion.
+    """
+
+    def __init__(self, alphabet: Sequence[str], num_cond_dim=0):
+        self.alphabet = alphabet
+        self.action_type_order = [GraphActionType.Stop, GraphActionType.AddNode]
+
+        self.num_tokens = len(alphabet) + 2  # Alphabet + BOS + PAD
```
maclandrol, line 91: forget my comment above.

src/gflownet/models/seq_transformer.py
```diff
@@ -0,0 +1,137 @@
...
+            state_preds,
+        )
+
+
+def generate_square_subsequent_mask(sz: int):
+    """Generates an upper-triangular matrix of -inf, with zeros on diag."""
+    return torch.triu(torch.ones(sz, sz) * float("-inf"), diagonal=1)
+
+
+class PositionalEncoding(nn.Module):
```
maclandrol, line 125: Consider newer (more efficient) PE

maclandrol (approved):

src/gflownet/envs/graph_building_env.py
bengioe: True, I'll add something

src/gflownet/envs/mol_building_env.py
bengioe: The check is done somewhere, but it would be more robust to not rely on that. Will add a check.

src/gflownet/models/seq_transformer.py
bengioe: I just reused Moksh's code here ^^"
Any suggestions on better implementations?

src/gflownet/envs/seq_building_env.py
bengioe: Yes this is only at the token level, all the model sees are token ids, a token could be a k-mer, although to properly "account for" backwards actions and different ways the same k-mer can be produced (with k additions vs 1 addition of the k-mer token) we'd need possibly non-trivial extra code.

src/gflownet/models/seq_transformer.py
maclandrol: If you care about very long sequences, then ALibi would make sense, otherwise, Rotary PE seems to work pretty well. See implementation of various PE here: https://github.com/lucidrains/x-transformers/blob/23f913aad5a93bff0a3c77267e04584a5f29f5ea/x_transformers/x_transformers.py#L215

bengioe: I added RotaryEmbedding, it looked interesting (but it would probably make sense to benchmark more of them at some point).

PR #107, opened 2023-09-06
Separate P_B trunk & min_len 

bengioe:
This PR makes it possible to have a separate set of parameters for P_B by setting `cfg.model.do_separate_p_b`. It also introduces `cfg.algo.min_len`, which defaults to `0` and prevents and agent from stopping before `min_len` steps have been taken.

Other changes:
- This change required adding a timestep parameter to `graph_to_Data` across contexts.
- While debugging I added a `consider_masks_complete` flag to contexts, which indicates when masks computed by a context should be taken as ground truth in terms of which actions are valid. This should help in some aspects in the future (e.g. counting parents, debugging new masks).
- Adds explicit mask computation in `SeqTransformerGFN`


bengioe:
Basic experiments on the double trunk model (2pPB) vs shared parameterized (pPB) vs uniform (uPB). The other parameter is `sampling_tau`
image

image

PR #110, opened 2023-10-23
Add backward sampling

bengioe:
This PR adds proper backward sampling routines, specifically to sample from parameterized backward policies or to sample uniformly (from masks).

This PR also:
- Changes how stem attachments are defined in the fragment environment using a saner "src_"/"dst_" format
- Includes #96 to showcase offline training
- adds an extra `Queue` object for in-main-process calls to an MP wrapped model
- changes `GraphActionCategorical._compute_batchwise_max` to use `finfo().min` instead of the batch min to compute the argmax -- this avoids some edge case errors.
- numpy-fies `graph_to_Data` in the fragment environment

Includes @diamondspark as an author since the PR is based on his implementation.

### DIFF
--- a/docs/examples/grid_cond_gfn.py
+++ b/docs/examples/grid_cond_gfn.py
@@ -253,7 +253,7 @@ class FlowNet_TBAgent:
         return log_ratio
 
     def learn_from(self, it, batch):
-        if type(batch) is list:
+        if isinstance(batch, list):
             log_ratio = torch.stack(batch, 0)
         else:
             log_ratio = batch
--- a/src/gflownet/envs/graph_building_env.py
+++ b/src/gflownet/envs/graph_building_env.py
@@ -606,7 +606,7 @@ class GraphActionCategorical:
         # there are no corresponding logits (this can happen if e.g. a
         # graph has no edges), we don't want to accidentally take the
         # max of that type, since we'd get 0.
-        min_val = torch.min(torch.stack([i.min() for i in x if i.numel()]))
+        min_val = torch.finfo().min
         outs = [torch.zeros(self.num_graphs, i.shape[1], device=self.dev) + min_val for i in x]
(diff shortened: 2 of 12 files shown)


src/gflownet/envs/graph_building_env.py
```diff
@@ -325,6 +330,32 @@ def reverse(self, g: Graph, ga: GraphAction):
...
+            return GraphAction(GraphActionType.AddEdge, source=ga.source, target=ga.target)
+        if ga.action == GraphActionType.RemoveNodeAttr:
+            return GraphAction(
+                GraphActionType.SetNodeAttr,
+                source=ga.source,
+                target=ga.target,
+                attr=ga.attr,
+                value=g.nodes[ga.source][ga.attr],
+            )
+        if ga.action == GraphActionType.RemoveEdgeAttr:
```
hadim: Why a serie of `if` instead of `if` + `elif` (and also an `else` as a sanity check)?

By doing so, you'll likely decrease the compute time spent doing all those checks.

docs/examples/grid_cond_gfn.py
```diff
@@ -253,7 +253,7 @@ def sample_many(self, mbsize):
         return log_ratio
 
     def learn_from(self, it, batch):
-        if type(batch) is list:
+        if isinstance(batch, list):
```
hadim, line 256: Clearly better than using `type`!

hadim (commented):
See a minor comment about the use of `if`.

src/gflownet/envs/graph_building_env.py
bengioe: Ah yes good catch

docs/examples/grid_cond_gfn.py
bengioe: You can thank `ruff` for catching this 😄

PR #116, opened 2024-02-06
add fixes from the maxent paper

SobhanMP:


src/gflownet/utils/conditioning.py
```diff
@@ -102,19 +104,20 @@ def sample(self, n):
...
+            m = Dirichlet(torch.FloatTensor([self.cfg.preference_param] * self.num_objectives))
             preferences = m.sample([n])
         else:
             raise ValueError(f"Unknown preference type {self.cfg.preference_type}")
         preferences = torch.as_tensor(preferences).float()
         return {"preferences": preferences, "encoding": self.encode(preferences)}
 
     def transform(self, cond_info: Dict[str, Tensor], flat_reward: Tensor) -> Tensor:
-        scalar_logreward = (flat_reward * cond_info["preferences"]).sum(1).clamp(min=1e-30).log()
+        # NO LOG NO LOG NO LOG NO LOG NO LOG NO LOG
```
bengioe: FYI we changed the way we handle this in #117 (which fixes this)

SobhanMP: I merged the main branch, will run tests tomorrow to check that everything is still working.

src/gflownet/models/graph_transformer.py
```diff
@@ -169,7 +169,7 @@ def __init__(
         self,
         env_ctx,
         cfg: Config,
-        num_graph_out=1,
+        num_graph_out=2,
```
bengioe: Is this intentional? May make more sense as a flag for models that output both F(s) and n(s)

src/gflownet/trainer.py
```diff
@@ -90,8 +99,13 @@ def compute_flat_rewards(self, mols: List[RDMol]) -> Tuple[FlatRewards, Tensor]:
...
 
 
+class Closable(Protocol):
+    def close(self):
+        pass
+
+
 class GFNTrainer:
-    def __init__(self, hps: Dict[str, Any]):
+    def __init__(self, hps: Dict[str, Any], print=True):
```
bengioe: Since this triggers printing the hyperparameters, perhaps rename to `print_hps`? Avoids confusion with `print_every`.

src/gflownet/data/qm9.py
```diff
@@ -4,59 +4,70 @@
...
+from gflownet.utils import sascore
+
 
 class QM9Dataset(Dataset):
-    def __init__(self, h5_file=None, xyz_file=None, train=True, target="gap", split_seed=142857, ratio=0.9):
+    def __init__(self, h5_file=None, xyz_file=None, train=True, targets=["gap"], split_seed=142857, ratio=0.9):
         if h5_file is not None:
             self.df = pd.HDFStore(h5_file, "r")["df"]
         elif xyz_file is not None:
             self.load_tar()
```
bengioe: Not sure if this is a mistake but this should probably take `xyz_file` as an argument

src/gflownet/tasks/qm9_moo.py
```diff
@@ -0,0 +1,375 @@
...
+def safe(f, x, default):
+    try:
+        return f(x)
+    except Exception:
+        return default
+
+
+class QM9GapMOOTask(QM9GapTask):
+    """Sets up a multiobjective task where the rewards are (functions of):
+    - the the binding energy of a molecule to Soluble Epoxide Hydrolases.
```
bengioe: I assume this should be homo-lumo gap

src/gflownet/tasks/qm9_moo.py
```diff
@@ -0,0 +1,375 @@
...
+        self.task = QM9GapMOOTask(
+            dataset=self.training_data,
+            cfg=self.cfg,
+            rng=self.rng,
+            wrap_model=self._wrap_for_mp,
+        )
+
+    def setup(self):
+        super().setup()
+        # self.sampling_hooks.append(
```
bengioe: Maybe add a flag or remove entirely

src/gflownet/trainer.py
```diff
@@ -344,31 +378,66 @@ def run(self, logger=None):
...
+        }
+        if self.sampling_model is not self.model:
+            state["sampling_model_state_dict"] = [self.sampling_model.state_dict()]
+        fn = pathlib.Path(self.cfg.log_dir) / "model_state.pt"
+        with open(fn, "wb") as fd:
+            torch.save(
+                state,
+                fd,
+            )
+        shutil.copy(fn, pathlib.Path(self.cfg.log_dir) / f"model_state_{it}.pt")
```
bengioe: Would be nice to have a flag here, saving every checkpoint can end up taking a lot of space so I wouldn't have it on by default.

src/gflownet/utils/metrics.py
```diff
@@ -522,6 +522,22 @@ def inv_transform(self, arr):
         return self.scale * arr + self.loc
 
 
+def chunkedsim(thresh, fp, mode_fps, delta=16):
```
bengioe: This doesn't seem to be called anywhere, although I can see how this is useful in a notebook.
I'd suggest a more expressive name, like `all_are_tanimoto_similar_to`

src/gflownet/utils/multiobjective_hooks.py
```diff
@@ -227,3 +235,27 @@ def finalize(self):
...
+        self.idx = idx
+
+    def __call__(self, trajs, rewards, flat_rewards, cond_info):
+        x = np.sort(flat_rewards.numpy(), axis=0)
+        ret = {}
+        y = np.sort(rewards.numpy())
+        for i, idx in enumerate(self.idx):
+            f = max(min(math.floor(x.shape[0] * idx), x.shape[0] - 1), 0)
+            for j in range(x.shape[1]):
+                ret[f"fr_{j}_{idx:.2f}%"] = x[f, j]
```
bengioe: Maybe prefixing this with `percentile` would make it more descriptive.

src/gflownet/trainer.py
```diff
@@ -129,7 +145,7 @@ def __init__(self, hps: Dict[str, Any]):
         # Print the loss every `self.print_every` iterations
         self.print_every = self.cfg.print_every
         # These hooks allow us to compute extra quantities when sampling data
-        self.sampling_hooks: List[Callable] = []
+        self.sampling_hooks: List[Callable] = [RewardStats()]
```
bengioe: Maybe add a comment saying this computes reward percentiles by default

pyproject.toml
```diff
@@ -90,6 +90,7 @@ dev = [
     "types-pkg_resources",
     # Security pin
     "gitpython>=3.1.30",
+    "ruamel.yaml",
```
bengioe: What is this for?

src/gflownet/data/qm9.py
SobhanMP: Nice catch!

src/gflownet/utils/metrics.py
SobhanMP: Sounds, good how about `all_are_tanimoto_different`?

SobhanMP: Also, this is mostly for speed, it's a fair bit faster on large mode_fps

pyproject.toml
SobhanMP: This is used in the qm9 file? `tasks/qm9/qm9.py`

SobhanMP: rewrote it with the default yaml implementation.

src/gflownet/trainer.py
SobhanMP: Removed it, unless you think it should be there, I don't think it should be there.

src/gflownet/trainer.py
SobhanMP: Sounds resonable.

src/gflownet/tasks/qm9_moo.py
SobhanMP: For performance reasons, would it make sense to remove this in both moo classes?

src/gflownet/models/graph_transformer.py
SobhanMP: This is intentional; I agree that it makes sense to handle it better. Are there any reasonable use cases where someone would want to have something higher than 2? 
Would it make sense to have a `dict[str,int]` as output to split the output into groups by default? So you may use something like
```
outsize = {'logF': 1, 'logN': 1}
```
And get a `dict` with those fields? While it's not type-checked, it feels more appropriate than doing the bookkeeping of what is what.

src/gflownet/trainer.py
SobhanMP: Switching to `yaml` causes the failure, I am not sure how to fix it. Does that example help? The yaml file seems very specific to recursion/valance. Should I remove it?

SobhanMP:
Just in case, please don't merge

src/gflownet/models/graph_transformer.py
bengioe: That's not a bad strategy, I'll make a note of this, but I'll think more about it. We will eventually need a strategy for n>1 per-state outputs that doesn't require always outputting max(all possible ns). For now I think just reading a flag in `StandardOnlineTrainer.setup_model` to do `+1` to `num_graph_out` would be enough.

src/gflownet/tasks/qm9_moo.py
bengioe: Considering we've reported those metrics in the MOGFN and GCGFN papers, we should leave them in the (seh MOO) code (at least as an option)

src/gflownet/trainer.py
bengioe: Is the second comment about `ruamel`/`yaml`? I guess we don't really need `qm9.yaml`, it's an artefact of older code that I never took the time to remove.

SobhanMP:
Having a generic moo class would be so nice.
BTW the moo task did not support having the FR in an order other than the default...

src/gflownet/online_trainer.py
```diff
@@ -43,6 +52,22 @@ def setup_data(self):
...
+        if self.cfg.opt.opt == "adam":
+            return torch.optim.Adam(
+                params,
+                lr,
+                (momentum, 0.999),
+                weight_decay=self.cfg.opt.weight_decay,
+                eps=self.cfg.opt.adam_eps,
+            )
+
+        raise NotImplementedError(f"{self.opt.opt} is not implemented")
```
bengioe: Should be `self.cfg.opt.opt`

SobhanMP: Oh you're right

SobhanMP:
It may be a stupid realization but `generate_forward_trajectory` is subsumed by `sample_backward_from_graphs` by setting `model=False`.

SobhanMP:
@bengioe is the `List[Dict[str, Any]]` ok in `trajectory_balance.py`? `mypy` complaints otherwise as the inferred type for the list is too restrictive.

SobhanMP:
things definitely run now

bengioe (approved):

PR #118, opened 2024-02-14
Add wandb-sweep example and clean-up use of configs

julienroyd:
The use of both hps-dictionaries in the Task files was a bit difficult to manage alongside the Config() dataclass objects that were now becoming the norm inside the repo. The mix of both would:

1. Make it difficult to keep-up in the task file with new configs and config name-changes in the Config objects since the dictionary structure needed to match the hierarchy in the config object without any direct dependency. IDE tools are much better at helping us track where a config is when using cfg.path.to.parameter than cfg{'path':{'to':{'parameter'}}}.
2. Lead to silent errors since assuming a name change on the config side, the corresponding parameter would be defined both on the user side and in Config(), and these two parameter-twins would be used in different places in the code without any error getting triggered. By forcing the user to provide a Config() object to the constructor, we enforce the correct hierarchy more easily.

Also added support for wandb:

- In the code but this is only optional and very light. In `GFNTrainer.log()` we simply check if a wandb.run is defined (whether wandb.init() has been called on the user side) and if so, we wandb.log() our metrics on the cloud too.
- Added a wandb-sweep example in `gflownet/hyperopt/wandb-demo'. Just 2 files to run (1) set up hyperparameters and wandb sweep (2) launch slurm workers to instantiate wandb-agents and work on the sweep.


julienroyd:
@bengioe, it seemed harmless to me, but while cleaning up the config-use if the Tasks' main(), I moved qm9.py back with the other tasks. Not sure why historically it was the only task who had its own directory and config file but we should revert that if for some reason that was important.

bengioe (approved):

PR #119, opened 2024-02-15
Update installation to python 3.10, torch 2.1

julienroyd:
A few updates regarding the installation setup:

- Updated torch and python versions to 2.1 and 3.10 respectively. torch-geometric, sparse etc still seem sensitive to specific versions so I kept them lucked but at least now we get the improvements from python 3.10 and torch 2.0.
- Added requirements/main-3.10.in and requirements/dev-3.10.in to regenerate the requirements/*.txt files used by `tox run`. The `.in` files are simply a copy-paste of the dependencies listed in the README. Not ideal since they need to be updated by hand and the .txt files also need to be regenerated by hand everytime we add a new core-dependency in `pyproject.toml`, but that doesn't happen very often and the user can still install the repo with a single pip install.
- Added wandb in dependencies (only optional to use -- it is activated outside of the main codebase by the user, but we need the package to wandb.log() the data inside the GFNTrainer() in case the user has activated it)
- Updated the README


julienroyd:
It was a bit of a battle but it is ready to go now.

bengioe (approved):

julienroyd:
Merging

PR #121, opened 2024-02-23
Minor fragment environment speed improvements + replay fix

bengioe:
Changes:
- Caches `Data` instance of graphs
- Compute the `remove_node_mask` more efficiently
- Remove the `deepcopy` in sampling iterator and fix a data overwriting bug
- `Data` cache of modified graphs is cleared in `step` with `Graph.clear_cache`


julienroyd:
LGTM, I'll just run a training with goal-conditioning to validate that we don't have issues with the replay buffer anymore.

julienroyd:
**Edit:** LGTM, the results seem to reproduce previous behavior (with deepcopies)
image

julienroyd (approved):

PR #122, opened 2024-03-01
Better iterators

### LINKED ISSUE #123: Nested dataclasses do not reinitialize
We use nested `dataclass`es in `Config`. This is generally find since we only instantiate a `Config` once, but in some cases (e.g. testing) way may want to initialize multiple. This will not work as expected since the inner instances belong to the `Config` class rather than its instances.

- [ ] replace `Inner()` config dataclasses by `field(default_factory=Inner)`

bengioe:
This PR refactors the `SamplingIterator` class into a `DataSource` class, where much of the existing functionality is decomposed into methods that can be combine with more clarity. Instead of interacting with `SamplingIterator` through a long list of sometimes unclear arguments, `DataSource` works by creating and combining iterators that perform more specific jobs.

Changes:
- `SamplingIterator` is now `DataSource`
- refactor around devices, device-needing code should now call `get_worker_device()`. This avoids passing around a device object (which, because of workers, can get confusing)
- adds an `is_eval` member to algorithms, and shifts the responsibility of determining random action probabilities to `GFNAlgorithm`
- adds `SQLiteLogHook` to generalize logging as a hook that's added to `DataSource` instances
- adds `AvgRewardHook`, a simple sampling hook to report average reward
- fixes minor MOO incongruities
- fixes #123 , whereby nested dataclasses would be only initialized once
- fixes a timeout condition typo in `MultiObjectiveStatsHook`

TODO:
- [x] address TODO items in this draft PR
- [x] complete transition to `get_worker_device` and remove device objects being passed around


src/gflownet/data/data_source.py
```diff
@@ -0,0 +1,315 @@
...
+                t = self.current_iter
+                p = self.algo.get_random_action_prob(t)
+                cond_info = self.task.sample_conditional_information(n_this_time, t)
+                # TODO: in the cond info refactor, pass the whole thing instead of just the encoding
+                trajs = self.algo.create_training_data_from_own_samples(model, n_this_time, cond_info["encoding"], p)
+                self.set_traj_cond_info(trajs, cond_info)  # Attach the cond info to the trajs
+                self.compute_properties(trajs, mark_as_online=True)
+                self.compute_log_rewards(trajs)
+                batch_info = self.call_sampling_hooks(trajs)
+                yield trajs, batch_info
```
julienroyd, line 131: Can we avoid code duplication here with do_sample_model() above? Or generalize do_sample_model_n_times to the n=1 case and use the same method for training and validation?

src/gflownet/trainer.py
```diff
@@ -199,103 +210,95 @@ def _wrap_for_mp(self, obj, send_to_device=False):
...
+        # Since the model may be used by a worker in a different process, we need to wrap it.
+        # The device `dev` returned here is the device that the worker will use to interact with the model;
+        # normally, if the main process has the model on 'cuda', this will simply be 'cpu' (since workers
+        # don't have CUDA access).
+        # See implementation_nodes.md for more details.
+        model, dev = self._wrap_for_mp(self.sampling_model, send_to_device=True)
+        replay_buffer, _ = self._wrap_for_mp(self.replay_buffer, send_to_device=False)
+
+        n_drawn = int(self.cfg.algo.global_batch_size * (1 - self.cfg.algo.offline_ratio))
+        n_replayed = n_drawn if self.cfg.replay.batch_size is None else self.cfg.replay.batch_size
```
julienroyd: I imagine we may want to mix it all, e.g. generate 16 trajs (keep in batch), sample 16 trajs from replay buffer and sample 32 trajs from the offline dataset. In that case I find it a bit hard to foresee how `replay.batch_size` and `replay.replaces_online_data` will interact (need to make sure that we don't bust `global_batch_size` when setting `replay.batch_size` etc). 

Would it be simpler to set only one config: `replay.batch_ratio` within `[0., 1.]`, which would define what proportion of the samples from the policy we want to allocate to the replay buffer v.s. online. 

e.g. something like:

```
    def build_training_data_loader(self) -> DataLoader:
        # Since the model may be used by a worker in a different process, we need to wrap it.
        # The device `dev` returned here is the device that the worker will use to interact with the model;
        # normally, if the main process has the model on 'cuda', this will simply be 'cpu' (since workers
        # don't have CUDA access).
        # See implementation_nodes.md for more details.
        model, dev = self._wrap_for_mp(self.sampling_model, send_to_device=True)
        replay_buffer, _ = self._wrap_for_mp(self.replay_buffer, send_to_device=False)

        n_from_dataset = int(self.cfg.algo.global_batch_size * self.cfg.algo.offline_ratio)
        n_from_policy = self.cfg.algo.global_batch_size - n_from_dataset
        n_replayed = int(n_from_policy * self.cfg.replay.batch_ratio)
        n_keep_in_batch = self.cfg.algo.global_batch_size - n_replayed

        src = DataSource(self.cfg, self.ctx, self.algo, self.task, replay_buffer=replay_buffer)
        if n_from_dataset:
            src.do_sample_dataset(self.training_data, n_from_dataset, backwards_model=model)
        if n_from_policy:
            src.do_sample_model(model, n_from_policy, n_keep_in_batch)
        if n_replayed and replay_buffer is not None:
            src.do_sample_replay(n_replayed)
        if self.cfg.log_dir:
            src.add_sampling_hook(SQLiteLogHook(str(pathlib.Path(self.cfg.log_dir) / "train"), self.ctx))
        for hook in self.sampling_hooks:
            src.add_sampling_hook(hook)
        return self._make_data_loader(src)
```

Then we could replace `keep_samples_in_batch` for `num_keep_in_batch`:
```
    def do_sample_model(self, model, num_samples, num_keep_in_batch):
        if num_keep_in_batch != num_samples:
            assert self.replay_buffer is not None, "Throwing away samples without a replay buffer"

        def iterator():
            while self.active:
                t = self.current_iter
                p = self.algo.get_random_action_prob(t)
                cond_info = self.task.sample_conditional_information(num_samples, t)
                # TODO: in the cond info refactor, pass the whole thing instead of just the encoding
                trajs = self.algo.create_training_data_from_own_samples(model, num_samples, cond_info["encoding"], p)
                self.set_traj_cond_info(trajs, cond_info)  # Attach the cond info to the trajs
                self.compute_properties(trajs, mark_as_online=True)
                self.compute_log_rewards(trajs)
                self.send_to_replay(trajs)  # This is a no-op if there is no replay buffer
                batch_info = self.call_sampling_hooks(trajs)
                yield (trajs[:num_keep_in_batch], batch_info[:num_keep_in_batch])

        self.iterators.append(iterator)
        return self
```

src/gflownet/trainer.py
```diff
@@ -199,103 +210,95 @@ def _wrap_for_mp(self, obj, send_to_device=False):
...
-            random_action_prob=0.0,
-            hindsight_ratio=0.0,
-            init_train_iter=self.cfg.num_training_steps,
-        )
+        model, dev = self._wrap_for_mp(self.model, send_to_device=True)
+        # TODO: we're changing the default, make sure anything that is using test data is adjusted
+        src = DataSource(self.cfg, self.ctx, self.algo, self.task, is_algo_eval=True)
+        n_drawn = int(self.cfg.algo.global_batch_size)
+
+        src = DataSource(self.cfg, self.ctx, self.algo, self.task, is_algo_eval=True)
```
julienroyd: Duplicate line `src= DataSource(...`

julienroyd (commented):
Taking a first look. Great refacto, I think it is much clearer now! Left a few comments (probably more to come).

src/gflownet/trainer.py
bengioe: Hmm, that's a good point. I wonder if it wouldn't be even simpler to just have these directly as four hyperparameters:
- `n_from_dataset`
- `n_from_policy`
- `n_replayed`
- `n_keep_in_batch`, Optional, and defaults to `n_from_policy` if unspecified (this would match the current behavior, but we could change it)

and then, if the users want to think of these as ratios of some global batch size, that's up to them.

I did all this calculus to not change the default behavior, but maybe it's better to "break" things (well, not really break as long as we change the tasks' `default_hps` accordingly) and also make them clearer.

src/gflownet/data/data_source.py
bengioe: Yes, definitely (`do_sample_model` is more like the n=infty case, but the comment stands).

src/gflownet/trainer.py
julienroyd: Yep I like that, I think that would be the cleanest way to go

julienroyd: Actually maybe just:

- `n_from_dataset`
- `n_from_policy`
- `n_from_replay`

We don't really need `n_keep_in_batch` right? Everything `from_policy` goes into the buffer and stays in the batch. If `n_from_replay > n_from_policy` we generate `n_from_replay` new samples at every iteration, put them in the buffer and simply keep `n_from_policy` in the batch (but the user doesn't have to think about it, they just chose what's in the batch with these 3 arguments).

bengioe: In our current implementation we override the policy samples with replay samples, i.e. `n_keep_in_batch` is 0. That's standard in RL, but afaik not standard in GFNs. Happy to drop `n_keep_in_batch` and change the default though (consider what GFN people seem to do).

julienroyd: Yeah I agree, I think most of the time people would exclusively use `from_policy` or `from_replay` or `from_dataset and from_policy`, but having the 3 settable in an independent way makes it easier to study the effect of online v.s. off-policy v.s. offline while keeping the config crystal clear.

bengioe: Following up on our discussion I ended going with:
- `algo.num_from_policy`
- `algo.num_from_dataset`
- `replay.num_from_replay`, Defaults to `algo.num_from_policy`, (i.e. 50/50)
- `replay.num_new_samples`, Defaults to `algo.num_from_policy`, if smaller than `num_from_policy` then not all on-policy samples will be added to the replay. If larger than `num_from_policy` then the training batch will not contain all the new samples, but the buffer will. For example, if one wishes to sample `N` samples every step but only add them to the buffer and not make them part of the training batch, then one should set `replay.num_new_samples=N` and `algo.num_from_policy`=0.

src/gflownet/data/data_source.py
bengioe: Actually, I'm thinking the two achieve slightly different purposes (one has replay interaction, the other doesn't; one is for training, the other for validation). The point of this setup being clarity, I'd be inclined to make each method as minimal as possible, even if there's going to be some repeated code.

src/gflownet/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        self, model: nn.Module, batch: gd.Batch, num_bootstrap: Optional[int] = 0
+    ) -> Tuple[Tensor, Dict[str, Tensor]]:
+        """Computes the loss for a batch of data, and proves logging informations
+
+        Parameters
+        ----------
+        model: nn.Module
+            The model being trained or evaluated
+        batch: gd.Batch
+            A batch of graphs
```
julienroyd, line 35: Naive question: does it have to be graph? Or was there sequence generation tasks also handled from this repo?

src/gflownet/algo/config.py
```diff
@@ -121,19 +126,20 @@ class AlgoConfig:
...
-    tb: TBConfig = TBConfig()
-    moql: MOQLConfig = MOQLConfig()
-    a2c: A2CConfig = A2CConfig()
-    fm: FMConfig = FMConfig()
-    sql: SQLConfig = SQLConfig()
+    tb: TBConfig = field(default_factory=TBConfig)
+    moql: MOQLConfig = field(default_factory=MOQLConfig)
+    a2c: A2CConfig = field(default_factory=A2CConfig)
+    fm: FMConfig = field(default_factory=FMConfig)
+    sql: SQLConfig = field(default_factory=SQLConfig)
```
julienroyd, line 145: How does this change the previous behavior?

src/gflownet/trainer.py
```diff
@@ -193,109 +122,97 @@ def setup(self):
...
+            src,
             batch_size=None,
             num_workers=self.cfg.num_workers,
             persistent_workers=self.cfg.num_workers > 0,
             prefetch_factor=1 if self.cfg.num_workers else None,
         )
 
+    def build_training_data_loader(self) -> DataLoader:
+        # Since the model may be used by a worker in a different process, we need to wrap it.
+        # See implementation_nodes.md for more details.
```
julienroyd: See implementation_notes.md *

src/gflownet/algo/config.py
```diff
@@ -95,8 +95,18 @@ class AlgoConfig:
...
-    global_batch_size : int
-        The batch size for training
+    num_from_policy : int
+        The number of on-policy samples for a training batch.
+        If using a replay buffer, see `replay.num_from_replay` for the number of samples from the replay buffer, and
+        `replay.num_new_samples` for the number of new samples to add to the replay buffer (e.g. `num_from_policy=0`,
+        and `num_new_samples=N` inserts `N` new samples in the replay buffer at each step, but does not make that data
+        part of the training batch).
+    num_from_dataset : int
+        The number of samples from the dataset for a training batch
```
julienroyd, line 105: super clear !

src/gflownet/data/data_source.py
```diff
@@ -0,0 +1,320 @@
...
+        self.task = task
+        self.replay_buffer = replay_buffer
+        self.is_algo_eval = is_algo_eval
+        self.sampling_hooks: List[Callable] = []
+        self.active = True
+
+        self.global_step_count = torch.zeros(1, dtype=torch.int64) + start_at_step
+        self.global_step_count.share_memory_()
+        self.global_step_count_lock = torch.multiprocessing.Lock()
+        self.current_iter = start_at_step
```
julienroyd, line 46: Correct me if I'm wrong, we don't resume stopped experiments yet, but this is meant to insure that we could if we implemented the checkpoint loading on the model & trainer sides?

src/gflownet/data/data_source.py
```diff
@@ -0,0 +1,320 @@
...
+                cond_info = self.task.sample_conditional_information(num_samples, t)
+                objs, props = map(list, zip(*[data[i] for i in idcs])) if len(idcs) else ([], [])
+                trajs = self.algo.create_training_data_from_graphs(objs, backwards_model, cond_info["encoding"], p)
+                self.set_traj_cond_info(trajs, cond_info)  # Attach the cond info to the trajs
+                self.set_traj_props(trajs, props)
+                self.compute_log_rewards(trajs)
+                yield trajs, {}
+
+        self.iterators.append(iterator)
+        return self
```
julienroyd, line 202: Less clear to me when this one is used (only for QM9 I believe?). Is it meant to compute the traj_likelihood of the model for these fixed examples?

julienroyd (approved):
Looks good to me! And much needed refacto, very nice result !

Small suggestion: could add just a few words about DataSource in implementation_notes.md, e.g.:

```
## Data sources

The data used for training GFlowNets can come from a variety of sources. `DataSource` implements these different use-cases as individual iterators that collectively assemble the training batches before passing it to the trainer. Some of these use-cases include:
- Generating new trajectories on-policy
- Sampling trajectories from passed policies from a replay buffer
- Sampling trajectories from a fixed, offline dataset 

`DataSource` also covers validation sets, including cases such as:
- Generating new trajectories (w.r.t a fixed dataset of conditioning goals)
- Evaluating the model's likelihood on trajectories from a fixed, offline dataset
```

src/gflownet/data/data_source.py
bengioe: Yes that's the idea

src/gflownet/data/data_source.py
bengioe: No this just creates the trajectories (backwards from a starting x that comes from a dataset) using either a uniform P_B, or `backwards_model` if provided. It's up to the Algo to compute likelihoods and such.

src/gflownet/algo/config.py
bengioe: If you were to create two configs, `a = Config()`, `b=Config()`, doing say `a.algo.max_len = 32` would change both `a` and `b`, because they'd both share `Config.algo == a.algo == b.algo`. This change reinstantiates those nested dataclasses properly.

src/gflownet/__init__.py
bengioe: No, that's true. For now sequences masquerade as graphs but it would probably be cleaner to generalize the code/update the documentation.

PR #124, opened 2024-03-09
add maxent GFNs

SobhanMP:
Will get back with some benchmarks.


src/gflownet/algo/trajectory_balance.py
```diff
@@ -613,6 +691,54 @@ def subtb_loss_fast(self, P_F, P_B, F, R, traj_lengths):
             total_loss[ep] = (F_start - F_end + P_F_sums - P_B_sums).pow(2).sum() / car[T]
```
bengioe: We should probably have `_loss` here as well for consistency.

bengioe: I went ahead and fixed it :)

bengioe (approved):

PR #125, opened 2024-03-11
Adding SC-Scorer

julienroyd:
Adding SCScorer in gflownet/models. This might be a more interesting metric / reward to use for synthetic accessibility than SA-score, see: https://pubs.acs.org/doi/abs/10.1021/acs.jcim.7b00622.

(not for immediate use but I had been playing around with it so I thought we might as well merge it in since this is very isolated from the rest of the codebase)


bengioe:
I'd rather not have models in the git repo itself, we could follow the pattern of `bengio2021flow.py`, and link to the original (or, in case it disappears, I made a copy here: https://storage.googleapis.com/emmanuel-data/models/connorcoley_scscore_model.ckpt-10654.as_numpy.pickle)

Other than that LGTM.

PR #129, opened 2024-03-28
Adds ActionIndex and Harmonize use of action-masks

julienroyd:
Masks were used somewhat ad-hoc here and there. Here we centralise masking in GraphActionCategorical(), specifically:

- renamed masks in GraphActionCategorical to action_mask (to avoid confusion with done-masking, molecule validity masking, goal-conditioning masking, etc.)
- made logits and masks private attributes of GraphActionCategorical
- remove _mask() application from GraphTransformerFragEnvelopeQL and GraphTransformerGFN
- removed masking from GraphSampler (masks applied in property setters of GraphActionCategorical._logits)

We also replaced the action-tuples `Tuple[int, int, int]` action type, row and column by a named-tuple `ActionIndex` to the code and typing more intuitive and readable.

### DIFF
--- a/src/gflownet/algo/graph_sampling.py
+++ b/src/gflownet/algo/graph_sampling.py
@@ -98,7 +98,7 @@ class GraphSampler:
         graphs = [self.env.new() for i in range(n)]
         done = [False] * n
         # TODO: instead of padding with Stop, we could have a virtual action whose probability
-        # always evaluates to 1. Presently, Stop should convert to a [0,0,0] aidx, which should
+        # always evaluates to 1. Presently, Stop should convert to a (0,0,0) ActionIndex, which should
         # always be at least a valid index, and will be masked out anyways -- but this isn't ideal.
         # Here we have to pad the backward actions with something, since the backward actions are
         # evaluated at s_{t+1} not s_t.
@@ -136,7 +136,7 @@ class GraphSampler:
                 actions = sample_cat.sample()
             else:
                 actions = fwd_cat.sample()
-            graph_actions = [self.ctx.aidx_to_GraphAction(g, a) for g, a in zip(torch_graphs, actions)]
+            graph_actions = [self.ctx.ActionIndex_to_GraphAction(g, a) for g, a in zip(torch_graphs, actions)]
             log_probs = fwd_cat.log_prob(actions)
             # Step each trajectory, and accumulate statistics
             for i, j in zip(not_done(range(n)), range(n)):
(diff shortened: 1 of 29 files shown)


julienroyd:
Can reproduce previous results with goal-conditioning and 4 objectives:
image

src/gflownet/envs/graph_building_env.py
```diff
@@ -696,35 +749,18 @@ def sample(self) -> List[Tuple[int, int, int]]:
         # mutually exclusive).
 
         # Uniform noise
-        u = [torch.rand(i.shape, device=self.dev) for i in self.logits]
+        u = [torch.rand(i.shape, device=self.dev) for i in self._masked_logits]
         # Gumbel noise
-        gumbel = [logit - (-noise.log()).log() for logit, noise in zip(self.logits, u)]
-
-        if self.masks is not None:
-            gumbel_safe = [
```
bengioe, line 704: This is removed because we're now using `-inf`?

src/gflownet/algo/graph_sampling.py
```diff
@@ -259,21 +256,17 @@ def not_done(lst):
             else:
                 gbatch = self.ctx.collate(torch_graphs)
                 action_types = self.ctx.bck_action_type_order
-                masks = [getattr(gbatch, i.mask_name) for i in action_types]
+                action_masks = [self.ctx.action_type_to_mask(t, gbatch, assert_mask_exists=True) for t in action_types]
```
bengioe: `action_type_to_mask` seems to be missing for `mol_building_env` and `seq_building_env`?

bengioe (approved):
LGTM, just two questions.

src/gflownet/envs/graph_building_env.py
julienroyd: Yes. I talked with Sobhan and he told me that he added that in because the multiplicative masking would not work if the model was unstable and outputting extremely small values for the logits. Masking with -inf should fix the issue.

src/gflownet/algo/graph_sampling.py
bengioe: Actually, I'm not sure this needs to be a context method (nor a graph model method), it could just be a function somewhere?

julienroyd: Yeah I agree, I wasn't sure where to put that. I can simply put it as a general function in `graph_building_env.py` for now?

bengioe: 👍  works for me.

PR #131, opened 2024-05-02
Add `raw_logits` in GraphActionCategorical to() method

bengioe:
`raw_logits` were not sent to the device, causing downstream errors depending on when the raw logits are used to compute the masked logits.

### DIFF
--- a/src/gflownet/envs/graph_building_env.py
+++ b/src/gflownet/envs/graph_building_env.py
@@ -627,6 +627,7 @@ class GraphActionCategorical:
 
     def to(self, device):
         self.dev = device
+        self.raw_logits = [i.to(device) for i in self.raw_logits]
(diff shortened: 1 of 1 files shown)


bengioe:
~Not sure why the Mac build fails, seems to be some bad config on github's side. Either way this is a minor fix and passes on Linux.~ Fixed by #132

bengioe:
@julienroyd things seem to work with this fix but I haven't extensively tested. Can you think of other places where we should check for a similar error?

julienroyd:
> @julienroyd things seem to work with this fix but I haven't extensively tested. Can you think of other places where we should check for a similar error?

No, I think it should be good, _masked_logits were already being sent to device so I think raw_logits were the only ones that were left behind.

julienroyd (approved):

PR #139, opened 2024-10-02
chore: modernize `get_version` and remove `types-pkg_resources`

dmaljovec:
update from `legion` on behalf of @dmaljovec

This will help prevent `roadie lock` issues stemming from usage of a "yanked" package


bengioe (approved):
Not sure why this particular notebook was changed, but it appears to just be formatting 👍 
LGTM

dmaljovec:
> Not sure why this particular notebook was changed, but it appears to just be formatting 👍 LGTM

Ah, sorry. the idea was to find and replace a block of code programmatically and then run `ruff` against the codebase. The script did not find the code it wanted to replace, and so there was no need to reformat any code here. I am updating the PR to be a more targeted change to hopefully make it more clear the actual intent of what we were trying to update here.

PR #150, opened 2025-07-04
[Docs] Visualizing gflownet's codebase via diagrams

ivanmilevtues:
I have generated high-level diagrams for the gflownet project. The idea is to help people get up-to-speed with the codebase you can see how it renders here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/gflownet/on_boarding.md

I have generated a diagram first documentation for gflownet. As a lot of scientists interact with this codebase, I am sure that they are rather interested in the main flow and method then reading the source code in details. That is the main idea of the diagrams they show the main components and how they interact with each other. Then one can dive deeper by clicking on a component of interest. This allow scientists to quickly modify and extend the project as they can focus only on the component of interest with the full codebase context.

Any feedback is more than welcome! We also just released a free GitHub Action that can automatically update the diagrams so they are always up-to-date.

If this sounds interesting I'd love to connect and chat if such thing can have more impact and how can we help the dev process for scientists. We already became part of the https://github.com/recursionpharma/EFAAR_benchmarking/pull/104.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.


ivanmilevtues:
Hey, a quick update on our side, this week we released our diagram generaiton engine as an open-source project! If you are interested to see more on how the generation works you can do so at https://github.com/CodeBoarding/CodeBoarding

PR #152, opened 2026-05-01
Fix atom backward mask for attributed edge removal

tgritsaev:
## Summary

This PR fixes a mismatch between the atom-level molecular environment's backward
action mask and the parent-counting logic used by the graph-building
environment.

Previously, `MolBuildingEnvContext.graph_to_Data` allowed `RemoveEdge` for any
non-bridge edge. This included edges with bond attributes such as
`type=BondType.DOUBLE`. However, `GraphBuildingEnv.parents()` and
`count_backward_transitions()` only allow direct edge removal when the edge has
no attributes; attributed edges should first remove the edge attribute and only
then remove the plain edge.

## Why this matters

When sampling uniform backward trajectories from QM9 molecules, the old mask
could sample `RemoveEdge` directly on a typed bond. The reverse forward action is
then only `AddEdge`, which does not recreate the original bond attribute. This
can produce inconsistent trajectories where applying the stored forward action
does not recover the stored child graph.

This mainly affects training with uniform backward and results in unstable training.

## Change

- Require direct `RemoveEdge` to be both non-bridge and attribute-free.
- Add a regression test with a three-atom cycle containing one typed bond.
- The test verifies that direct removal of the typed bond is masked out, while
  removing the bond attribute remains allowed, and that the backward mask count
  agrees with `count_backward_transitions`.

## Validation

Locally verified the regression logic with an RDKit-enabled environment:

```text
remove_edge_mask 0.0
remove_edge_attr_mask 1.0
mask_count 3.0
count_backward_transitions 3
```

Also verified on a QM9 backward-sampling diagnostic that the fix removes the
inconsistent attributed-edge transitions:

```text
before: bad_trajs 21 / 256, bad_steps 21 / 3002
after:  bad_trajs 0, bad_steps 0, count_mismatches 0
```

### DIFF
--- a/src/gflownet/envs/mol_building_env.py
+++ b/src/gflownet/envs/mol_building_env.py
@@ -326,7 +326,7 @@ class MolBuildingEnvContext(GraphBuildingEnvContext):
 
         remove_edge_mask = np.zeros((len(g.edges), 1), dtype=np.float32)
         for i, e in enumerate(g.edges):
-            if e not in bridges:
+            if e not in bridges and len(g.edges[e]) == 0:
                 remove_edge_mask[i] = 1
 
         edge_attr = np.zeros((len(g.edges) * 2, self.num_edge_dim), dtype=np.float32)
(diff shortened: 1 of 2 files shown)


bengioe (approved):
Good catch!
Thank you for the contribution.

